PR #11, opened 2024-12-04
Feature/populate

cristianoCatolico:
It populates tenants table reading from kickstart copy in core service and also validates tenantId and UserID in the token with fusionAuth.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -43,11 +43,12 @@ func main() {
 	// Handlers
 	authHandlers := handlers.NewAuthHandlers(authService)
 	targetHandlers := handlers.NewTargetHandlers(targetService)
+	tenantHandlers := handlers.NewTenantHandlers(tenantService)
 	result, err := utils.OpenAndReadKickstartJson(tenantService)
 	fmt.Println(result)
 
 	// Server
-	s := api.NewAPIServer(":8000", targetHandlers, authHandlers)
+	s := api.NewAPIServer(":8000", targetHandlers, tenantHandlers, authHandlers)
 
 	if err := s.Init(); err != nil {
 		log.Fatalf("Failed to initialize APIServer: `%+v`", err)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,7 +8,7 @@ import (
 	"github.com/kptm-tools/core-service/pkg/handlers"
 	"github.com/kptm-tools/core-service/pkg/services"
 	"github.com/kptm-tools/core-service/pkg/storage"
-	//"github.com/kptm-tools/core-service/pkg/utils"
+	"github.com/kptm-tools/core-service/pkg/utils"
 )
 
 func main() {
@@ -37,12 +37,13 @@ func main() {
 	// Services
 	authService := services.NewAuthService()
 	targetService := services.NewTargetService(coreStore)
-	//tenantService := services.NewTenantService(coreStore)
+	tenantService := services.NewTenantService(coreStore)
 
 	// Handlers
 	authHandlers := handlers.NewAuthHandlers(authService)
 	targetHandlers := handlers.NewTargetHandlers(targetService)
-	//utils.openAndReadKickstartJson(tenantService)
+	result, err := utils.openAndReadKickstartJson(tenantService)
+	log.Fatalf("Failed to initialize APIServer");
 	// Server
 	s := api.NewAPIServer(":8000", targetHandlers, authHandlers)
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"log"
 
 	"github.com/kptm-tools/core-service/pkg/api"
@@ -42,8 +43,9 @@ func main() {
 	// Handlers
 	authHandlers := handlers.NewAuthHandlers(authService)
 	targetHandlers := handlers.NewTargetHandlers(targetService)
-	result, err := utils.openAndReadKickstartJson(tenantService)
-	log.Fatalf("Failed to initialize APIServer");
+	result, err := utils.OpenAndReadKickstartJson(tenantService)
+	fmt.Println(result)
+
 	// Server
 	s := api.NewAPIServer(":8000", targetHandlers, authHandlers)
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,6 +8,7 @@ import (
 	"github.com/kptm-tools/core-service/pkg/handlers"
 	"github.com/kptm-tools/core-service/pkg/services"
 	"github.com/kptm-tools/core-service/pkg/storage"
+	//"github.com/kptm-tools/core-service/pkg/utils"
 )
 
 func main() {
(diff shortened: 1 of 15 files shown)


pkg/api/auth.go
```diff
@@ -82,12 +87,28 @@ func WithAuth(endpoint http.HandlerFunc, functionName string) http.HandlerFunc {
...
-			var roles = token.Claims.(jwt.MapClaims)["roles"]
+			var tenantID = token.Claims.(jwt.MapClaims)["tid"]
+			log.Print(tenantID)
+			var userID = token.Claims.(jwt.MapClaims)["sub"]
+			log.Print(userID)
+			// Build request
+			req, err := buildFusionAuthVerifyRequest(tenantID.(string), userID.(string))
+
+			if err != nil {
+				fmt.Errorf("Invalid request build")
```
grez-lucas: `fmt.Errorf()` is a function that creates a custom Go Error with the message you provided, in this case I suggest returning the error message as a:

 `return WriteJSON(w, http.StatusInternalServerError, APIError{ Error: err.Error() }` 

to handle  the case were building the FusionAuthVerifyRequest did not work.

pkg/api/auth.go
grez-lucas, line 1: While this approach (validating with a client that connects to fusionAuth) works (and its OK!), I tried something similar on my Login service approach to perform requests to FusionAuth and ended up switching to the [FusionAuth Go Client](https://github.com/FusionAuth/go-client) package to handle responses in a cleaner way. I suggest we add a task to refactor this section of the code to use that package after this sprint.

pkg/utils/populate.go
```diff
@@ -0,0 +1,59 @@
...
+		if err != nil {
+			fmt.Println(err)
+		}
+		fmt.Println(resultDB)
+
+	}
+	return "Good", nil
+}
+
+func buildFusionAuthGetTenantsRequest() (*http.Request, error) {
```
grez-lucas: I don't think this method is used anywhere in the codebase. Lets **get rid of it for now**, since it can be replaced with the [FusionAuth Go Client](https://github.com/FusionAuth/go-client) `client.RetrieveTenants` method if we wish to integrate it somewhere.

pkg/utils/parser.go
grez-lucas, line 1: Great work with this parser!

pkg/utils/populate.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func OpenAndReadKickstartJson(tenantService interfaces.ITenantService) (string, error) {
+	dir, err := os.Getwd()
+	if err != nil {
+		fmt.Println("Error getting current working directory:", err)
+		return "Can not access to path", err
+	}
+	fmt.Println("Current working directory:", dir)
+	jsonFile, err := os.Open(dir + "/pkg/utils/fusionauth/kickstart/kickstart.json")
```
grez-lucas, line 17: Wouldn't we want to open the parent repository kickstart json file?

pkg/utils/fusionauth/kickstart/kickstart.json
grez-lucas, line 1: While this is **OK** **for testing**, it involves duplicating our already existing `kickstart.json` file from the [Parent Repository](https://github.com/kptm-tools/kptm-pentest/tree/dev/utils/fusionauth/kickstart). 

A cleaner approach would be to reference that instead in our parser, but I see difficulties in a Dockerized environment...

Lets **discuss this tomorrow**.

cmd/main.go
```diff
@@ -36,13 +38,17 @@ func main() {
 	// Services
 	authService := services.NewAuthService()
 	targetService := services.NewTargetService(coreStore)
+	tenantService := services.NewTenantService(coreStore)
 
 	// Handlers
 	authHandlers := handlers.NewAuthHandlers(authService)
 	targetHandlers := handlers.NewTargetHandlers(targetService)
+	tenantHandlers := handlers.NewTenantHandlers(tenantService)
+	result, err := utils.OpenAndReadKickstartJson(tenantService)
```
grez-lucas, line 47: This leads to behavior where each time `core-service` restarts, the JSON file is re-opened, parsed and the db is re-populated. I suggest you integrate this `OpenAndReadKickstartJson` method in some way that allows it to run only when our core-service DB is first created and configured. (Maybe `storage/storage.go` is the best place for this).

image

pkg/api/auth.go
```diff
@@ -51,15 +55,16 @@ func WithAuth(endpoint http.HandlerFunc, functionName string) http.HandlerFunc {
...
 				checkAudience := token.Claims.(jwt.MapClaims).VerifyAudience(aud, false)
 
 				if !checkAudience {
 					return nil, fmt.Errorf("Invalid audience")
 				}
+				*/
 
 				// verify iss claim: Make sure the issuer is as expected
-				iss := "http://localhost:9011" // TODO: Set this to env var
+				iss := "https://piedpipervideochat.com" // TODO: Set this to env var
```
grez-lucas: Lets keep this issuer claim verification hardcoded to "https://app.kriptome.com" for now, which is what our **blueprint application** will use.

pkg/api/auth.go
```diff
@@ -82,12 +87,28 @@ func WithAuth(endpoint http.HandlerFunc, functionName string) http.HandlerFunc {
...
+				WriteJSON(w, http.StatusUnauthorized, APIError{Error: "Not able to verify with AuthProvider"})
+			}
+			if resp.StatusCode != 200 {
+				WriteJSON(w, http.StatusUnauthorized, APIError{Error: "User not authorized in the given Tenant"})
+			}
 
+			var roles = token.Claims.(jwt.MapClaims)["roles"]
 			parsedRoles, err := domain.GetRolesFromStringSlice([]string{roles.([]interface{})[0].(string)})
 
 			if err != nil {
```
grez-lucas, line 113: Missing return statement after this WriteJSON call, which leads to this funny behavior (lol):

image

pkg/storage/tenant.go
grez-lucas, line 1: Nice work!

grez-lucas (changes requested):
Please fix unexpected behaviors on auth JWT decoding, and parsing the `kickstart` file multiple times. Don't worry about integrating the `FusionAuth Go Client` package yet (unless you wish to and feel it might help speed things up.) Lets keep things functional.

For formatting and linting, I suggest using the [VSCODE Official Go Extension](https://code.visualstudio.com/docs/languages/go).

Alternatively, you can use the `Makefile` in our repository root and use: 
- `make build` to format, lint and build binaries
- `make run`  to format, lint, build binaries and run said binaries locally
- `make test` to run go tests locally

pkg/utils/populate.go
cristianoCatolico: Ok I remove

pkg/api/auth.go
cristianoCatolico: Ok I change

pkg/storage/tenant.go
cristianoCatolico: Thanks

pkg/utils/parser.go
cristianoCatolico: Thanks

pkg/api/auth.go
cristianoCatolico: I see thanks for the explanation

pkg/utils/populate.go
cristianoCatolico: Yes, I will try to be the parent directory

pkg/api/auth.go
cristianoCatolico: I agree with you

cmd/main.go
cristianoCatolico: I have create function to verify if exist in order to avoid duplication with tenantID

pkg/api/auth.go
cristianoCatolico: Ok got it, thanks Lucas

pkg/utils/fusionauth/kickstart/kickstart.json
cristianoCatolico: I like also the cleaner approach. I willy to have it.

grez-lucas (approved):

PR #18, opened 2024-12-11
Refactor/kt 39 auth quality assurance

grez-lucas:
Refactored `middleware/auth` to follow cleaner coding practices, and incorporated unit tests into our application specific logic. 

- The HUGE `withAuth` method was broken down into smaller, more granular components which are easier to test and debug. 
- Specific error handling was incorporated as well, following the [Sentinel error pattern](https://bitfieldconsulting.com/posts/wrapping-errors).
     - This error handling pattern is slightly different than what is currently incorporated in `services/auth`. It provides a cleaner way of testing, wrapping and checking for custom error types (using `errors.Is()` instead of `errors.As()`, which is why `services/auth` and `handlers/auth` should be refactored as well in the future to follow this pattern.

- Now `withAuth` doesn't write a JSON response with the specific error, this logging is left to the `core-service`. It is a good idea to keep Unauthorized responses as ambiguous as possible, to prevent unauthorized individuals from knowing why the system is failing to authorize them ;)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -50,11 +50,11 @@ func (s *APIServer) Init() error {
 	// Auth routes
 	router.HandleFunc("POST /api/login", makeHTTPHandlerFunc(s.authHandlers.Login))
 	router.HandleFunc("POST /api/tenant", makeHTTPHandlerFunc(s.authHandlers.RegisterTenant))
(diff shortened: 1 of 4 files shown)


cristianoCatolico (approved):
Clean code, clean test, thank you very much for this refactor

PR #20, opened 2024-12-12
[feat]: adding endpoint forgot password

cristianoCatolico:
This endpoint receives an email and applicationId and send the email with the link to change password. It lefts to do the endpoint of change password. This pr also contains the definitions for endpoints of registration however I could not make it work yet.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -54,6 +54,7 @@ func (s *APIServer) Init() error {
 	// Auth routes
 	router.HandleFunc("POST /api/login", makeHTTPHandlerFunc(s.authHandlers.Login))
 	router.HandleFunc("POST /api/forgot-password", makeHTTPHandlerFunc(s.authHandlers.ForgotPassword))
+	router.HandleFunc("POST /api/change-password", makeHTTPHandlerFunc(s.authHandlers.ChangePassword))
 	router.HandleFunc("POST /api/user", makeHTTPHandlerFunc(s.authHandlers.RegisterUser))
 	router.HandleFunc("POST /api/user/{id}/verify-email", makeHTTPHandlerFunc(s.authHandlers.VerifyEmail))
 	router.HandleFunc("POST /api/tenant", makeHTTPHandlerFunc(s.authHandlers.RegisterTenant))
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -53,6 +53,9 @@ func (s *APIServer) Init() error {
 
 	// Auth routes
 	router.HandleFunc("POST /api/login", makeHTTPHandlerFunc(s.authHandlers.Login))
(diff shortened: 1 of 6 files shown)


pkg/api/api.go
```diff
@@ -53,6 +53,9 @@ func (s *APIServer) Init() error {
 
 	// Auth routes
 	router.HandleFunc("POST /api/login", makeHTTPHandlerFunc(s.authHandlers.Login))
+	router.HandleFunc("POST /api/forgot-password", makeHTTPHandlerFunc(s.authHandlers.ForgotPassword))
+	router.HandleFunc("POST /api/register-user", makeHTTPHandlerFunc(s.authHandlers.RegisterUser))
```
grez-lucas: Regarding these endpoints, I feel we could make them more RESTful by instead using:

- `POST /api/register-user` -> `POST /api/user`
- `POST /api/verify-email` -> `POST /api/user/{id}/verify-email`

pkg/handlers/auth.go
grez-lucas, line 1: This is on the right track! I imagine the forgotPasswordRequest is being reused for these handlers but will be switched later on

pkg/handlers/requests.go
grez-lucas, line 1: Nice

pkg/api/api.go
cristianoCatolico: Thanks I will change it.

pkg/handlers/requests.go
cristianoCatolico: Thanks

pkg/handlers/auth.go
cristianoCatolico: I dit not get the idea, but I will annotate to ask later.

pkg/services/auth.go
```diff
@@ -364,3 +364,130 @@ func (s *AuthService) NewFusionAuthClient() (*fusionauth.FusionAuthClient, error
...
+
+	userFusion, faErr, err := client.RetrieveUser(userID)
+
+	if err != nil {
+		return nil, err
+	}
+	if faErr != nil {
+		return nil, NewFaError(userFusion.StatusCode, faErr.Error())
+	}
+	if userFusion.User.Verified {
```
grez-lucas, line 477: Nice error handling!

grez-lucas (approved):
Nice PR! You're slowly becoming a GoLang expert 😎. Make sure to document these new endpoints and Request/Response objects in Apiary when you get the chance.

I expect some of the logic for the `POST /api/tenant` endpoint to require some refactors from my side after this PR, since there are new tenant and application configurations we're taking into account. 

Don't forget to commit your changes to the `kickstart.example.json` (which should be a copy of the `kickstart.json` file) **in the parent repository**. As a note here:
- I merged a PR recently where I hard-coded the Tenant1's UUID and App1 UUID, make sure your new `kickstart.example.json` includes these variable definitions ;)

PR #25, opened 2024-12-20
Host management

cristianoCatolico:


pkg/domain/host.go
```diff
@@ -8,33 +8,52 @@ import (
...
-	Value      string    `json:"host_value"`
-	Type       HostType  `json:"host_type"`
-	CreatedAt  time.Time `json:"created_at"`
-	UpdatedAt  time.Time `json:"updated_at"`
+	ID          string    `json:"id"`
+	TenantID    string    `json:"tenant_id"`
+	OperatorID  string    `json:"user_id"`
+	Name        string    `json:"name"`
+	Domain      string    `json:"domain"`
+	Ip          string    `json:"ip"`
```
grez-lucas: struct field Ip should be IP [ST1003](https://staticcheck.dev/docs/checks#ST1003)

pkg/domain/host.go
```diff
@@ -8,33 +8,52 @@ import (
...
+	Rapporteurs []byte    `json:"rapporteurs"`
+	CreatedAt   time.Time `json:"created_at"`
+	UpdatedAt   time.Time `json:"updated_at"`
+}
+
+type HostResponse struct {
+	ID          string       `json:"id"`
+	Name        string       `json:"name"`
+	Domain      string       `json:"domain"`
+	Ip          string       `json:"ip"`
```
grez-lucas: struct field Ip should be IP [ST1003](https://staticcheck.dev/docs/checks#ST1003)

pkg/services/host.go
```diff
@@ -22,35 +26,73 @@ func (s *HostService) CreateHost(t *domain.Host) (*domain.Host, error) {
...
+	isDeleted, err := s.storage.DeleteHostByID(ID)
+
+	if err != nil {
+		return false, err
+	}
+
+	return isDeleted, nil
+}
+
+func (s *HostService) GetHostname(ip_port string) string {
```
grez-lucas: Should not use underscores in Go names; method parameter ip_port should be ipPort

pkg/storage/host.go
```diff
@@ -87,6 +90,88 @@ func (s *PostgreSQLStore) GetHostsByTenantID(tenantID string) ([]*domain.Host, e
...
+		&host.Ip,
+		&host.Name,
+		&host.Credentials,
+		&host.Rapporteurs,
+		&host.CreatedAt,
+		&host.UpdatedAt)
+
+	switch err {
+	case sql.ErrNoRows:
+		return nil, fmt.Errorf("No rows were returned: `%+v`", err)
```
grez-lucas: error strings should not be capitalized [ST1005](https://staticcheck.dev/docs/checks#ST1005)

pkg/storage/host.go
```diff
@@ -87,6 +90,88 @@ func (s *PostgreSQLStore) GetHostsByTenantID(tenantID string) ([]*domain.Host, e
...
```
grez-lucas: error strings should not be capitalized [ST1005](https://staticcheck.dev/docs/checks#ST1005)

pkg/handlers/host.go
```diff
@@ -35,32 +38,52 @@ func (h *HostHandlers) CreateHost(w http.ResponseWriter, req *http.Request) erro
...
+	}
+
+	return api.WriteJSON(w, http.StatusCreated, constructResponse(host))
+}
 
-	host := domain.NewHost(createHostRequest.Value, t, createHostRequest.TenantID, createHostRequest.OperatorID, createHostRequest.Name)
+func (h *HostHandlers) GetHostsByTenantIDAndUserID(w http.ResponseWriter, req *http.Request) error {
 
-	host, err := h.hostService.CreateHost(host)
+	tenantID := req.Context().Value(middleware.ContextTenantID).(string)
```
grez-lucas, line 52: Nice use of context here!

go.mod
```diff
@@ -10,3 +10,8 @@ require (
 require github.com/FusionAuth/go-client v0.0.0-20241126020005-1254a936741f
 
 require github.com/google/uuid v1.6.0
+
+require (
+	github.com/jpillora/go-tld v1.2.1 // indirect
+	golang.org/x/net v0.0.0-20220225172249-27dd8689420f // indirect
```
grez-lucas: I suggest we use `golang.org/x/net v0.33.0` since this version you're using has multiple vulnerabilities. 


To update this package to it's stable and most secure version, run `go get golang.org/x/net/idna@v0.33.0`, and that should update this package automatically :)

pkg/api/api.go
grez-lucas, line 1: Thanks for cleaning out these endpoints to use the /api prefix! Also good job making use of the function identifier for role protection ;)

I must admit we should be **consistent** with our api naming convention. If we're using plural (like `api/hosts/{id}` instead of `api/host/{id}`, we should stick with that convention for ALL other endpoints. I agree that keeping things _plural_ makes more sense in a RESTful sense. This means refactoring:

* `/api/tenant/` ➡️ `api/tenants/`
* `api/user/` ➡️ `api/users/` 
* ...

And notifying @diegocarhuavilca 🔔

pkg/api/api.go
```diff
@@ -60,8 +60,11 @@ func (s *APIServer) Init() error {
 	router.HandleFunc("POST /api/tenant", makeHTTPHandlerFunc(s.authHandlers.RegisterTenant))
 	router.HandleFunc("GET /api/user/{id}", middleware.WithAuth(makeHTTPHandlerFunc(s.authHandlers.GetUser), "getUser"))
 
-	router.HandleFunc("POST /hosts", makeHTTPHandlerFunc(s.hostHandlers.CreateHost))
-	router.HandleFunc("GET /hosts", makeHTTPHandlerFunc(s.hostHandlers.GetHostsByTenantID))
+	router.HandleFunc("POST /api/hosts", middleware.WithAuth(makeHTTPHandlerFunc(s.hostHandlers.CreateHost), "newHost"))
+	router.HandleFunc("GET /api/hosts", middleware.WithAuth(makeHTTPHandlerFunc(s.hostHandlers.GetHostsByTenantIDAndUserID), "getHostsByTenantAndUser"))
+	router.HandleFunc("GET /api/hosts/{id}", middleware.WithAuth(makeHTTPHandlerFunc(s.hostHandlers.GetHostByID), "getHostByID"))
+	router.HandleFunc("DELETE /api/hosts/{id}", middleware.WithAuth(makeHTTPHandlerFunc(s.hostHandlers.DeleteHostByID), "deleteHostByID"))
+	router.HandleFunc("PATCH /api/hosts/{id}", middleware.WithAuth(makeHTTPHandlerFunc(s.hostHandlers.DeleteHostByID), "patchHostByID"))
```
grez-lucas: I'll assume this is just a temporary placeholder for the PATCH request handler ;)

pkg/domain/host.go
```diff
@@ -8,33 +8,52 @@ import (
...
+type Credential struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
 
-const (
-	Domain HostType = "domain"
-	IP     HostType = "ip"
-)
+type Rapporteur struct {
```
grez-lucas, line 14: Rapporteurs should only have a `Name`, `Email` and `IsPrincipal` attribute, no password. This may have been a typo on your side 🤣

```suggestion
type Rapporteur struct {
    Name string `json:"name"`
    Email string `json:"email"`
    IsPrincipal bool `json"is_principal"`
```

pkg/domain/host.go
grez-lucas, line 1: # Big review incoming here 👾

I see you went with the []byte design approach for storing these structures **directly in JSON format into the DB**. While this current approach works for an MVP, it's crucial to evaluate the scalability and security implications early. 

This is acceptable, since the system is small-scale and the data in the JSON fields won't grow significantly in the short term. Regardless, we should keep this flexible enough to accommodate changes in the future, which is why I suggest the following hybrid approach:

## 1. Keep the JSON approach for the `Rapporteur` object
* To store rapporteurs as JSON, use the `JSONB` data type in PostgreSQL instead of a byte array for better query performance and indexing:
```sql
ALTER TABLE hosts
ADD COLUMN credentials JSONB,
ADD COLUMN rapporteurs JSONB;
```
## 2. Normalize credentials, given their **sensitivity and importance for security.** 

For example:
```sql
CREATE TABLE credentials (
    id SERIAL PRIMARY KEY,
    host_id UUID REFERENCES hosts(id),
    username TEXT,
    password TEXT
);
```

## 3. For the `Host` object, keep attributes in line with our objects, **leave decoding-encoding logic in our storage.**
```go
type Host struct {
    ID          string       `json:"id"`
    TenantID    string       `json:"tenant_id"`
    OperatorID  string       `json:"user_id"`
    Name        string       `json:"name"`
    Domain      string       `json:"domain"`
    IP          string       `json:"ip"`
    Credentials []Credential `json:"credentials"`
    Rapporteurs []Rapporteur `json:"rapporteurs"`
    CreatedAt   time.Time    `json:"created_at"`
    UpdatedAt   time.Time    `json:"updated_at"`
}
```
* Modify the `NewHost()` method to accomodate these refactors
* leave decoding-encoding logic in our storage, this means marshalling/unmarshalling our objects into bytes at the moment of fetching them or creating them in our `pkg/storage/host.go` logic.
* There is no need to declare another storage file like `pkg/storage/credentials.go` (unless you really want to). You could keep SQL queries that perform `JOIN` operations within the `pkg/storage/host.go` logic.

pkg/handlers/host.go
```diff
@@ -69,12 +92,74 @@ func (h *HostHandlers) GetHostsByTenantID(w http.ResponseWriter, req *http.Reque
...
+	result := make(map[string]string)
+	if isDeleted {
+		result["deleted"] = "true"
+	} else {
+		result["deleted"] = "false"
+	}
+	return api.WriteJSON(w, http.StatusOK, result)
+}
+
+func constructHostForDB(createHostRequest *CreateHostRequest, req *http.Request, h *HostHandlers) *domain.Host {
```
grez-lucas, line 145: This sounds like a method that should be contained in our `storage` package, not the `handlers` package.

pkg/handlers/host.go
```diff
@@ -69,12 +92,74 @@ func (h *HostHandlers) GetHostsByTenantID(w http.ResponseWriter, req *http.Reque
...
+	hostResponse.CreatedAt = host.CreatedAt
+	hostResponse.ID = host.ID
+	hostResponse.Domain = host.Domain
+	hostResponse.Ip = host.Ip
+	json.Unmarshal(host.Credentials, &hostResponse.Credentials)
+	json.Unmarshal(host.Rapporteurs, &hostResponse.Rapporteurs)
+	return hostResponse
+}
+
+func getDomainIPValues(createHostRequest *CreateHostRequest, h *HostHandlers) (string, string) {
```
grez-lucas, line 167: Please correct me if I'm wrong, but I see you're using this method to **fill out the IP attribute** within a host, which is something we should not be doing. We want our users to either pass a Domain OR an IP, and leave it there. 

Later we want to check if this Domain OR IP is valid, which can be done via ping request to either the Domain or the IP in question. 

Additionally, this function should return an error if it fails. You're ignoring that error in line 153

pkg/handlers/requests.go
```diff
@@ -1,10 +1,23 @@
...
+import "github.com/kptm-tools/core-service/pkg/domain"
+
+type CredentialRequest struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
+
+type RapporteurRequest struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
```
grez-lucas: Again, Rapporteurs should have Name, Email and IsPrinciple, no Password.

pkg/handlers/requests.go
grez-lucas, line 1: Are `CredentialRequest` and `RapporteurRequest` planned to be used in the future? In what context?

pkg/interfaces/host.go
```diff
@@ -8,10 +8,17 @@ import (
 
 type IHostService interface {
 	CreateHost(*domain.Host) (*domain.Host, error)
-	GetHostsByTenantID(tenantID string) ([]*domain.Host, error)
+	GetHostsByTenantIDAndUserID(tenantID string, userID string) ([]*domain.Host, error)
+	GetHostByID(ID string) (*domain.Host, error)
+	GetHostname(string) string
+	DeleteHostByID(ID string) (bool, error)
+	PatchHostByID(ID, domainName, ip, alias string, credential, rapporteur []byte) (*domain.Host, error)
```
grez-lucas: With the refactor I suggested, this PatchHostByID method would receive an `ID` and `domain.Host` parameter and return `(*domain.Host, error)`

pkg/middleware/auth.go
grez-lucas, line 1: Nice use of context ;)

pkg/services/host.go
```diff
@@ -22,35 +26,63 @@
...
+	var timout time.Duration = 2
+	conf := &tls.Config{
+		InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
grez-lucas: Please give this a look 😬

pkg/services/host.go
grez-lucas: Again, it seems like you're using this GetHostname method to find out the Domain associated to an IP. On it's own it's useful, but if you're using it to fill out the IP, that's not in scope for our system just yet. I suggest applying security fixes but not getting rid of this function.

pkg/storage/host.go
```diff
@@ -3,17 +3,20 @@ package storage
...
 
 func (s *PostgreSQLStore) CreateHostsTable() error {
 	query := `create table if not exists hosts (
       id SERIAL PRIMARY KEY,
       tenant_id UUID,
       operator_id UUID,
-      value VARCHAR(2048) UNIQUE NOT NULL,
-      type VARCHAR(10) NOT NULL,
+      domain VARCHAR(2048) UNIQUE NOT NULL,
+      ip VARCHAR(15) NOT NULL,
```
grez-lucas: Since we may or may not have data on the IP-Domain (we could have one or the other) I suggest removing the NOT NULL for now.

pkg/storage/host.go
```diff
@@ -3,17 +3,20 @@ package storage
...
+      alias VARCHAR(2048) UNIQUE NOT NULL,
+      credentials	JSONB,
```
grez-lucas: The `JSONB` approach is super!

pkg/storage/host.go
grez-lucas, line 1: Overall pretty clean implementation! Please implement the hybrid approach (only for `Credentials`), so that we may follow security guidelines in the way we store sensitive data.

grez-lucas (changes requested):
Good first approach to this User Story. The biggest concern is regarding the JSONB vs Normalized DB storage implementation for our data structures. Please solve comments and let me know if anything needs further clarification. 🤠

pkg/handlers/host.go
cristianoCatolico: Thanks

go.mod
cristianoCatolico: Thanks for the observation

pkg/api/api.go
cristianoCatolico: Yes for sure :D

pkg/middleware/auth.go
cristianoCatolico: Thanks

pkg/api/api.go
cristianoCatolico: Ok I will notify Diego

pkg/domain/host.go
cristianoCatolico: Yes, it was my mistake, I corrected it.

pkg/domain/host.go
cristianoCatolico: Thanks for the observation

pkg/storage/host.go
cristianoCatolico: Ok, I removed it

pkg/storage/host.go
cristianoCatolico: Thanks

pkg/storage/host.go
cristianoCatolico: Thanks

pkg/services/host.go
cristianoCatolico: yes you are right, correcting

pkg/domain/host.go
cristianoCatolico: done

pkg/handlers/host.go
cristianoCatolico: It remains in the code but not enable in the api.

pkg/handlers/requests.go
cristianoCatolico: It is deleted

pkg/handlers/requests.go
cristianoCatolico: It is deleted this file

pkg/services/host.go
cristianoCatolico: Remains in code but not enable in api

pkg/storage/host.go
cristianoCatolico: We are encrypting the password. So we are covering it.

pkg/storage/host.go
cristianoCatolico: Yes

pkg/domain/host.go
cristianoCatolico: Refactor done

pkg/storage/host.go
```diff
@@ -77,33 +136,234 @@
...
+	queryInside := ""
+	for _, data := range credentialObject {
+		// Validate that ID is not empty for update otherwise does not update that record
+		if data.ID != "" {
+			queryInside += "(" + data.ID + ",'" + data.Username + "',pgp_sym_encrypt('" + data.Password + "', 'MAMA', 'compress-algo=1, cipher-algo=aes256')),"
+		}
+	}
+	queryInside = strings.TrimSuffix(queryInside, ",")
+	query := fmt.Sprintf("UPDATE credentials SET username = data.name, password=data.pass FROM (VALUES %s ) AS data(id, name, pass) WHERE credentials.id = data.id;", queryInside)
+	_, err := s.db.Query(query)
```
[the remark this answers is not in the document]
grez-lucas: +1 to copilot's suggestion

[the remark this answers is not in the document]
grez-lucas: This helps us avoid SQL injection! See the [Avoiding SQL Injection when inserting records into a postgresql database](https://www.calhoun.io/inserting-records-into-a-postgresql-database-with-gos-database-sql-package/) for more details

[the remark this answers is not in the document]
grez-lucas: @cristianoCatolico

grez-lucas (approved):
Good work! 😎

PR #26, opened 2024-12-31
Feature/kt 146 create event bus

grez-lucas:
Importing common package defintions to implement NATS event bus into core-service


cristianoCatolico (approved):
Simple and clear integration of nats. Thanks Lucas

PR #27, opened 2025-01-03
Integration of Event Bus into core-service

cristianoCatolico:


pkg/handlers/scan.go
```diff
@@ -0,0 +1,54 @@
...
+		var mr *malformedRequest
+
+		if errors.As(err, &mr) {
+			return api.WriteJSON(w, mr.status, api.APIError{Error: mr.Error()})
+		} else {
+			return api.WriteJSON(w, http.StatusInternalServerError, api.APIError{Error: err.Error()})
+		}
+	}
+
+	scan, err := s.scanService.CreateScans(scanRequest.HostIds)
```
grez-lucas, line 40: This err value is never used

pkg/services/scan.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, error
+	}
+	return dataScan, nil
+}
+
+func launchEvent(s interfaces.IStorage, ch chan resultChannel, wg *sync.WaitGroup, hostID string) {
+	defer wg.Done()
+
+	if hostID == "" {
+		ch <- resultChannel{err: fmt.Sprintf("error: got HostID empty")}
```
grez-lucas: unnecessary use of fmt.Sprintf

pkg/services/scan.go
grez-lucas, line 1: It's awesome that you ventured into trying out concurrency! There are a couple of small improvements that can be made in this file regarding error handling, concurrency and readability, which I'll try to leave as suggestions 😉

pkg/services/scan.go
```diff
@@ -0,0 +1,106 @@
...
+		storage: storage,
+	}
+}
+
+type resultChannel struct {
+	data domain.Host
+	err  string // you must use error type here
+}
+
+func (s ScanService) CreateScans(hostIDs []string) (*domain.Scan, error) {
```
grez-lucas, line 29: The `CreateScans` function is large and has several responsibilities (launch events, result aggregation, metadata creation, and storage interaction). A good readability improvement would be to break down this function into smaller helper functions, e.g:

- `launchScanEvents(hostIDs []string) ([]resultChannel, error)`: Encapsulates logic for launching goroutines and collecting results.
- `launchEvent(s interfaces.IStorage, ch chan resultChannel, wg *sync.WaitGroup, hostID string)`: Launches a single event and sends the result to the results channel 
- `addScanResult(scanDB *domain.Scan, host domain.Host)`: Adds the result of a single host scan to the Scan object.
- `createMetadata() []*domain.Metadata`: Initializes metadata for default services.
- `createTarget(host domain.Host) *events.Target`: Handles the creation of the events.Target.
- `createHostStatus(host domain.host, metadata []*domain.Metadata) *domain.StatusHost`:  Encapsulates the creation of StatusHost.

Just as an idea!

pkg/services/scan.go
```diff
@@ -0,0 +1,106 @@
...
+
+func NewScanService(storage interfaces.IStorage) *ScanService {
+	return &ScanService{
+		storage: storage,
+	}
+}
+
+type resultChannel struct {
+	data domain.Host
+	err  string // you must use error type here
```
grez-lucas: I think this comment is right lol. If the attribute `err` is of type `error` you could handle errors in a more idiomatic way when working with this channel. E.g:

```go
            if hostID == "" {
                ch <- resultChannel{err: fmt.Errorf("hostID is empty")}
                return
            }
```

Instead of passing a string 🤔

pkg/domain/scan.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/google/uuid"
+	res "github.com/kptm-tools/common/common/events"
+	"time"
+)
+
+type Metadata struct {
+	Progress string `json:"progress"`
+	Service  string `json:"service"`
```
grez-lucas: This Metadata struct is a good idea! The Service being an `string` gives me bad code smell though since it encourages _hardcoding strings_. 

I think it would be nice if our `common` package includes an `Enum` or string constants of type `ServiceType`, which we can reference throughout the codebase instead of hardcoding strings like `"Whois"`, `"Harvester"`, `"DNSLookup"`. 

For example:

Common Package Definition:
```go
// ServiceName represents the name of a service.
type ServiceName string

const (
    ServiceWhois      ServiceName = "Whois"
    ServiceHarvester  ServiceName = "Harvester"
    ServiceDNSLookup  ServiceName = "DNSLookup"
)
```

Then in our core-service code:
```go
metadataWhois := &domain.Metadata{
    Progress: "0%",
    Service:  string(ServiceWhois),
}
metadataHarvester := &domain.Metadata{
    Progress: "0%",
    Service:  string(ServiceHarvester),
}
metadataDNSLookup := &domain.Metadata{
    Progress: "0%",
    Service:  string(ServiceDNSLookup),
}
```

This would allow us to improve our EventResult definitions later on as well

pkg/handlers/scan.go
```diff
@@ -0,0 +1,54 @@
...
+	scanStartedPayload := &cmmn.ScanStartedEvent{
+		ScanID:    scan.ID,
+		Targets:   scan.Targets,
+		Timestamp: scan.CreatedAt.Unix(),
+	}
+	scanStartedString, err := json.Marshal(scanStartedPayload)
```
grez-lucas: My bad here for giving you poor example code. Maybe I'm being a bit pedantic, but scanStartedString is of type []byte. So the correct naming of this variable should be:
```suggestion
	scanStartedBytes, err := json.Marshal(scanStartedPayload)
```

pkg/services/scan.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func launchEvent(s interfaces.IStorage, ch chan resultChannel, wg *sync.WaitGroup, hostID string) {
+	defer wg.Done()
+
+	if hostID == "" {
+		ch <- resultChannel{err: fmt.Sprintf("error: got HostID empty")}
+		return
+	}
+	host, err := s.GetHostByID(hostID)
```
grez-lucas, line 58: If you change your result channel `err` attribute, you could do some awesome **Error Wrapping** stuff here. Error wrapping allows us to _wrap_ an already existing error (like a db error in this case) and give it more context. Here is a good visualization:

image


```suggestion
	host, err := s.GetHostByID(hostID)
	if err != nil { 
	ch <- resultChannel{err: fmt.Errorf("failed to get host: %w", err)}
	   return
	   }
```

The `%w` in the previous code allows you to easily wrap an error ;)

pkg/services/scan.go
```diff
@@ -0,0 +1,106 @@
...
+	data domain.Host
+	err  string // you must use error type here
+}
+
+func (s ScanService) CreateScans(hostIDs []string) (*domain.Scan, error) {
+	scanDB := domain.NewScan()
+
+	var wg sync.WaitGroup
+	var results []resultChannel
+	ch := make(chan resultChannel)
```
grez-lucas: To improve concurrency and prevent synchronous blocking, you can use a buffered channel here.

```suggestion
	ch := make(chan resultChannel, len(hostIDs)) // buffered channel
```

grez-lucas (changes requested):
Good work! This code is _functional_, there are a couple of improvements that can be made regarding **Error Handling**, **Concurrency** and **Readability** in the ` services/scan.go` file. Please give these a look today and let me know if you need any help 😁

pkg/domain/scan.go
grez-lucas: I'll implement the Common Package Definition right now and notify you when it's done!

grez-lucas: Done! To get access to these variables, run `go get github.com/kptm-tools/common@v1.2.5` ✅

pkg/handlers/scan.go
cristianoCatolico: Corrected

pkg/services/scan.go
cristianoCatolico: Thanks

pkg/services/scan.go
cristianoCatolico: Yes you are right, thanks for this observation 👍

pkg/services/scan.go
cristianoCatolico: Thanks teacher, I did not know that

pkg/handlers/scan.go
cristianoCatolico: Corrected

pkg/services/scan.go
cristianoCatolico: Corrected

pkg/domain/scan.go
cristianoCatolico: Done!, thanks Lucas

pkg/services/scan.go
cristianoCatolico: Almost all the idea I added

pkg/services/scan.go
cristianoCatolico: Thanks for the example and clarification

grez-lucas (approved):
Looks solid! Good work 😎👍

PR #28, opened 2025-01-06
fix(scan) Fix scan service bug, remove unnecessary concurrency

grez-lucas:
* Fixed an error where `domain.Scan` struct had unintended null values on `Targets` when creating a scan, which led to a ScanStartedEvent failing
* Cleaned up concurrency   

Concurrency is often unnecessary when interacting with a database. Concurrency can add complexity without meaningful performance gains, which was the case here, since it was leading to unintended errors. It's best to leave concurrency for when we discover bottlenecks in our system.


cristianoCatolico:
I thought that doing concurrent we gain some time, but I did not realize that could case a problem. New learning for me, thanks Lucas.

PR #40, opened 2025-01-10
Validation of hostname

cristianoCatolico:


pkg/handlers/host.go
```diff
@@ -159,20 +159,18 @@ func (h *HostHandlers) ValidateHost(w http.ResponseWriter, req *http.Request) er
...
-	var validation string
-	var err error
-	if len(validateHostRequest.Value) > 0 {
-		validation, err = h.hostService.ValidateHost(validateHostRequest.Value)
-	} else {
-		validation, err = h.hostService.ValidateAlias(validateHostRequest.Hostname)
-	}
 
-	if err != nil {
+	_, errValue := h.hostService.ValidateHost(validateHostRequest.Value)
```
grez-lucas, line 163: This ignores cases where ValidateHost pings the host and doesn't get a response

grez-lucas: E.g: 

```json
{
    "value": "https://www.i2linkedddddd.com"
}
```

Returns 200 OK, when it should return 404 Bad request

grez-lucas (changes requested):

PR #43, opened 2025-01-14
Redo scan create

cristianoCatolico:


pkg/domain/scan.go
```diff
@@ -20,19 +19,22 @@ type StatusHost struct {
...
 	Targets      []events.Target `json:"targets,omitempty"`
-	CreatedAt    time.Time       `json:"created_at"`
-	UpdatedAt    time.Time       `json:"updated_at"`
+	StartedAt    time.Time       `json:"started_at"`
+	EndedAt      time.Time       `json:"ended_at"`
 }
 
 func NewScan() *Scan {
 	return &Scan{
-		ID:        uuid.NewString(),
```
grez-lucas, line 35: Does this mean we're not going with string ID's for scans anymore? What is the reasoning behind this? I believe we have defined Scans having a string UUID everywhere else in the codebase 🤣

pkg/domain/scan.go
```diff
@@ -20,19 +19,22 @@ type StatusHost struct {
...
-		CreatedAt: time.Now().UTC(),
```
grez-lucas, line 36: I understand the creation of the StartedAt and EndedAt attributes, but replacing them with these two doesn't make too much sense:

This is the way I see it - 
CreatedAt: when the object was stored in our DB
UpdatedAt: when the object was last updated in our DB
StartedAt: when the scan was started
EndedAt: when the scan was ended

I still feel each of these are valuable from a DB perspective, so I don't think rushing to delete the first two is the right approach. Instead, we should leverage these, and make use of the NewScan function to declare attributes such as `CreatedAt` and `UpdatedAt`, while leaving `EndedAt` with it's default [zero-value](https://yourbasic.org/golang/default-zero-value/)

pkg/domain/scan.go
```diff
@@ -20,19 +19,22 @@ type StatusHost struct {
...
-		UpdatedAt: time.Now().UTC(),
+		StartedAt: time.Now().UTC(),
+		EndedAt:   time.Now().UTC(),
```
grez-lucas: It would be less confusing if we leave `EndedAt` as it's [zero-value](https://yourbasic.org/golang/default-zero-value/) on creation 
```suggestion
		return &Scan{
		    ID: uuid.NewString(),
		    StartedAt: time.Now().UTC(), // If you're posting a ScanStartedEvent when you create an instance, else it might not make too much sense
		    CreatedAt: time.Now().UTC(),
		    UpdatedAt: time.Now().UTC(),
```

pkg/storage/scan.go
```diff
@@ -40,52 +77,73 @@ func (s *PostgreSQLStore) ClearScansTable() error {
...
+    INSERT INTO scans_hosts (scan_id, host_id,status)
+    values ($1, $2, $3)`
 
-	scan := new(domain.Scan)
-	cols, _ := rows.Columns()
-	columns := make([]interface{}, len(cols))
-	columnPointers := make([]interface{}, len(cols))
-	for i := range columns {
-		columnPointers[i] = &columns[i]
+	for _, hostId := range sc.HostIDs {
```
grez-lucas: [ST1003](https://staticcheck.dev/docs/checks#ST1003) range var hostId should be hostID

pkg/api/api.go
grez-lucas, line 1: I think you forgot to define the roles for `getScans` (RoleOperator, RoleAnalyst). I had to do some tinkering to test out this endpoint, please add that.

pkg/storage/scan.go
```diff
@@ -1,19 +1,57 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+
+}
+
+func (s *PostgreSQLStore) CreateScanHostsTable() error {
+	query := `create table if not exists scans_hosts (
```
grez-lucas: the name `scans_hosts` should be `scan_hosts`, since it is a table used for associating a scan with multiple hosts. The idea of this table is to **normalize** our DB schema.

grez-lucas: `scan_hosts` should have a much simpler schema, this table is only for normalizing our db entities. 
```suggestion
query := `create table if not exists scan_hosts (
    id SERIAL PRIMARY KEY,
    scan_id INTEGER REFERENCES scans(id) ON DELETE CASCADE,
    host_id INTEGER REFERENCES hosts(id) ON DELETE CASCADE
)
```

pkg/storage/scan.go
```diff
@@ -1,19 +1,57 @@
...
+func (s *PostgreSQLStore) CreateScanTable() error {
 	query := `create table if not exists scans (
-      id UUID PRIMARY KEY,
-      status JSONB,
-      results JSONB,
+      id SERIAL PRIMARY KEY,
+      tenant_id UUID,
+      operator_id UUID,
+      started_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+      ended_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
```
grez-lucas: Scan status should go on this table, not on `scan_hosts`

```suggestion
     status VARCHAR(50) NOT NULL, -- e.g., 'pending', 'in_progress', 'completed', 'failed'
```

pkg/storage/scan.go
```diff
@@ -1,19 +1,57 @@
...
+	return nil
+
+}
+
+func (s *PostgreSQLStore) CreateScanHostsTable() error {
+	query := `create table if not exists scans_hosts (
+      id SERIAL PRIMARY KEY,
+      scan_id INT REFERENCES scans(id) ON DELETE CASCADE,
+      host_id INT REFERENCES hosts(id) ON DELETE CASCADE,
+      severity JSONB,
```
grez-lucas: `severity` and `vulnerability` should go in `scan_results`, which is a table for the results of a scan performed on multiple hosts. For vulnerabilities, parsing them through the `result` column of the nmap result could suffice. The key here is to write the correct `scan_results` table.

pkg/storage/scan.go
```diff
@@ -1,19 +1,57 @@
...
+	_, err := s.db.Query(query)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *PostgreSQLStore) CreateScanResultsTable() error {
```
grez-lucas, line 57: As mentioned before, this table is key. If we're working with many hosts per scan, I suggest we add a `host_id` column to our table to differentiate which result is assigned to who.

```sql
CREATE TABLE scan_results (
    id SERIAL PRIMARY KEY,
    scan_id INTEGER REFERENCES scans(id) ON DELETE CASCADE,
    host_id INTEGER REFERENCES hosts(id) ON DELETE CASCADE,
    tool_name VARCHAR(255) NOT NULL,
    result JSONB, -- Tool output for the specific host
    status VARCHAR(50) NOT NULL, -- e.g., 'pending', 'success', 'failed'
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);
```

Right now, our `Nmap` tool is the one that posts vulnerabilities, so storing it's output as JSONB like we do here and then parsing it using json.Unmarshal seems like a possibility.

If we wanted to go even further, we could define a table for `tools` with a schema like this:

```sql
CREATE TABLE tools (
    id SERIAL PRIMARY KEY,
    name VARCHAR(255) UNIQUE NOT NULL, -- Tool name
    description TEXT,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);
```

Then reference it here via:

`tool_id INTEGER REFERENCES tools(id) ON DELETE CASCADE,`

grez-lucas: ## Example Workflow
### When a Scan is Created

 1. Add an entry to the scans table.
 2. Add entries to the scan_hosts table for each host.
 3. For each tool and each host, initialize rows in the scan_results table with a pending status.

### When a Tool Completes

- Update the scan_results table for the specific scan_id, host_id, and tool_id with the tool's result and set the status to success.

### When the Scan is Complete
- Update the status column in the scans table to completed.

pkg/storage/scan.go
```diff
@@ -40,52 +77,73 @@ func (s *PostgreSQLStore) ClearScansTable() error {
...
-    INSERT INTO scans (id, status, created_at, updated_at)
+    INSERT INTO scans ( tenant_id, operator_id, started_at, ended_at)
     values ($1, $2, $3, $4)
-    RETURNING id, created_at, updated_at`
+    RETURNING *`
 
-	rows, err := s.db.Query(query, sc.ID, status, sc.CreatedAt, sc.UpdatedAt)
-	if err != nil {
-		return nil, fmt.Errorf("error creating scan: `%v`", err)
+	row := tx.QueryRow(query, sc.TenantID, sc.OperatorID, sc.StartedAt, sc.EndedAt)
```
grez-lucas: Nice addition of transactions ;)

pkg/storage/scan.go
```diff
@@ -1,19 +1,57 @@
...
 )
 
-func (s *PostgreSQLStore) CreateScansTable() error {
+func (s *PostgreSQLStore) CreateScanTable() error {
 	query := `create table if not exists scans (
-      id UUID PRIMARY KEY,
-      status JSONB,
-      results JSONB,
+      id SERIAL PRIMARY KEY,
+      tenant_id UUID,
```
grez-lucas: The `tenant_id`, `operator_id`, and `host_id` fields lack `NOT NULL` constraints in their respective tables. This can lead to data integrity issues.

pkg/storage/scan.go
```diff
@@ -40,52 +77,73 @@ func (s *PostgreSQLStore) ClearScansTable() error {
...
+	}
+	defer tx.Rollback()
 
-	status, _ := json.Marshal(sc.HostsStatus)
 	query := `
-    INSERT INTO scans (id, status, created_at, updated_at)
+    INSERT INTO scans ( tenant_id, operator_id, started_at, ended_at)
     values ($1, $2, $3, $4)
-    RETURNING id, created_at, updated_at`
+    RETURNING *`
```
grez-lucas: The use of `RETURNING *` is risky because it retrieves all columns without specifying them explicitly. This can lead to maintenance issues if table schemas change.

pkg/storage/scan.go
```diff
@@ -40,52 +77,73 @@ func (s *PostgreSQLStore) ClearScansTable() error {
...
+		}
 	}
+	return nil
+}
 
-	if err := rows.Scan(columnPointers...); err != nil {
-		return nil, err
+func scanIntoScan(row *sql.Row, scan *domain.Scan) error {
+	if err := row.Scan(&scan.ID, &scan.TenantID, &scan.OperatorID, &scan.StartedAt, &scan.EndedAt); err != nil {
+		return fmt.Errorf("error scanning rows: %w", err)
```
grez-lucas: The error message uses "error scanning rows," but the function deals with a single `row`. It should reflect the correct context.

pkg/storage/scan.go
```diff
@@ -40,52 +77,73 @@ func (s *PostgreSQLStore) ClearScansTable() error {
...
 
-	scan := new(domain.Scan)
-	cols, _ := rows.Columns()
-	columns := make([]interface{}, len(cols))
-	columnPointers := make([]interface{}, len(cols))
-	for i := range columns {
-		columnPointers[i] = &columns[i]
+	for _, hostId := range sc.HostIDs {
+		if _, err := tx.Exec(query, sc.ID, hostId, 0); err != nil {
+			return fmt.Errorf("failed to insert credential: %w", err)
```
grez-lucas: failed to insert credential lol

pkg/storage/scan.go
grez-lucas: If `sc.HostIDs` is empty, this method silently succeeds, which may not be desirable. Explicitly handle this case to avoid confusion.

pkg/storage/scan.go
grez-lucas, line 1: ## General Observations

* No logging is present to track failures or unexpected behavior during database operations, which makes debugging in production difficult.

grez-lucas (changes requested):
Good first approach! Please consider the schema suggestions, you have done a good job with this first version, and I feel we're getting closer to a clean implementation of storing scan data in our async system

pkg/storage/scan.go
cristianoCatolico: corrected

cristianoCatolico: Ok Lucas

pkg/storage/scan.go
cristianoCatolico: oh no what terrible mistake, too much copy paste

pkg/storage/scan.go
cristianoCatolico: True, corrected

pkg/storage/scan.go
cristianoCatolico: Thanks to you

pkg/api/api.go
cristianoCatolico: Yes your are right sorry, I added now

pkg/storage/scan.go
cristianoCatolico: Ok I added

pkg/storage/scan.go
cristianoCatolico: Ok I did not consider that, thank you

pkg/domain/scan.go
cristianoCatolico: You are right, I deleted that field

pkg/storage/scan.go
cristianoCatolico: corrected

cristianoCatolico: Ok Lucas, I will changed it

pkg/storage/scan.go
cristianoCatolico: ok I got it

pkg/domain/scan.go
cristianoCatolico: Ok Lucas

pkg/storage/scan.go
cristianoCatolico: ok Lucas

pkg/storage/scan.go
cristianoCatolico: When Scan is created works, but when tool and scan completes no

pkg/domain/scan.go
```diff
@@ -28,13 +28,32 @@ type Scan struct {
...
 	HostsResults []ResultHost    `json:"hosts_results,omitempty"`
 	Targets      []events.Target `json:"targets,omitempty"`
+	CreatedAt    time.Time       `json:"created_at,omitempty"`
+	UpdatedAt    time.Time       `json:"updated_at,omitempty"`
 	StartedAt    time.Time       `json:"started_at"`
 	EndedAt      time.Time       `json:"ended_at"`
+	Status       string          `json:"status,omitempty"`
+}
+
+type ScanSummary struct {
```
grez-lucas, line 45: This struct clears things up much more! Just a question: I'm assuming the `Severity` attribute is this.

image
 
In which case, this attribute being an array of integers is not very descriptive, and requires external context to interpret (e.g., which index represents "CRITICAL," "HIGH," etc.)

I suggest the following approach, to make our backend lives easier when it comes to filling out these attributes later on:

```suggestion
type SeverityCounts struct {
	Critical int `json:"critical"`
	High     int `json:"high"`
	Medium   int `json:"medium"`
	Low      int `json:"low"`
}


type ScanSummary struct {
ScanDate      string         `json:"scan_date,omitempty"`
	Host          string         `json:"host,omitempty"`
	Vulnerabilities int          `json:"vulnerabilities,omitempty"`
	Severities    SeverityCounts `json:"severities,omitempty"`
	Duration      int            `json:"duration,omitempty"`
	Status        int            `json:"status,omitempty"`
}

```

pkg/domain/scan.go
```diff
@@ -28,13 +28,32 @@ type Scan struct {
...
+	ScanDate      string `json:"scan_date,omitempty"`
+	Host          string `json:"host,omitempty"`
+	Vulnerability int    `json:"vulnerability,omitempty"`
```
grez-lucas: What does the `Vulnerability` attribute mean in this context? I'm having trouble interpreting it, since a Scan should detect many vulnerabilities 🧐

pkg/domain/scan.go
```diff
@@ -28,13 +28,32 @@ type Scan struct {
...
+	Severity      []int  `json:"severity,omitempty"`
+	Duration      int    `json:"duration,omitempty"`
+	Status        int    `json:"status,omitempty"`
+}
+
+type Tool struct {
```
grez-lucas, line 61: Nice!

pkg/storage/scan.go
```diff
@@ -65,6 +67,73 @@ func (s *PostgreSQLStore) CreateScanResultsTable() error {
...
+	_, err := s.db.Query(query)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *PostgreSQLStore) InsertTools() error {
```
grez-lucas, line 95: Nice use of the common module!

pkg/storage/scan.go
```diff
@@ -65,6 +67,73 @@ func (s *PostgreSQLStore) CreateScanResultsTable() error {
...
+	toolsData := s.getDefaultTools()
+	tx, err := s.db.Begin()
+	if err != nil {
```
grez-lucas, line 98: And nice error handling

pkg/storage/scan.go
```diff
@@ -84,9 +153,9 @@ func (s *PostgreSQLStore) CreateScan(sc *domain.Scan) (*domain.Scan, error) {
 	defer tx.Rollback()
 
 	query := `
-    INSERT INTO scans ( tenant_id, operator_id, started_at, ended_at)
-    values ($1, $2, $3, $4)
-    RETURNING *`
+    INSERT INTO scans ( tenant_id, operator_id,status, started_at, ended_at)
+    values ($1, $2,'PENDING', $3, $4)
+    RETURNING id, tenant_id, operator_id, status, started_at, ended_at`
```
grez-lucas: Great!

grez-lucas (commented):
Very nice refactor :1st_place_medal: please review the suggestions in `pkg/domain/scan.go`, then we should be good! I look forward to seeing the GetScans implementation ;)

pkg/domain/scan.go
cristianoCatolico: Good approach for severities

pkg/storage/scan.go
cristianoCatolico: For next feature

pkg/handlers/scan.go
```diff
@@ -61,18 +62,31 @@ func (s ScanHandlers) CreateScans(w http.ResponseWriter, req *http.Request) erro
 
 		return api.WriteJSON(w, http.StatusInternalServerError, err.Error())
 	}
+	for _, dataScan := range scans {
```
grez-lucas, line 65: Nice!

pkg/domain/scan.go
grez-lucas, line 1: This looks very good @cristianoCatolico 😁! There are just a couple of suggestions regarding consistency with ID's (particularly scanIDs). 

Because we're developing a distributed system, it would be ideal for our ScanID to be of type UUID, which is what you're sort of defining in line 
```go
type Scan struct {
	ID           string         `json:"id,omitempty"`
```

Except it's later contradicted in references such as

```go
type ScanResult struct {
	ScanID int
```

UUID's align with our event-driven architecture, since scans and results are processed asynchronously across components. They ensure uniqueness and improve security.

In PostgreSQL, you can store UUIDs efficiently using the native `UUID` data type, and modern libraries in Go (like `github.com/google/uuid`) make UUID generation easy.


## For example:

### Domain stuff

```go
type Scan struct {
	ID        uuid.UUID `json:"id"
	Name      string    `json:"name"`
	CreatedAt int64     `json:"created_at"`
}

func NewScan() *Scan {
	return &Scan{
		ID:        uuid.New(), // Generate the UUID here before inserting to DB
		StartedAt: time.Now().UTC(),
		CreatedAt: time.Now().UTC(),
		UpdatedAt: time.Now().UTC(),

	}
}
```


### Our schema would look something like this for `scans`

```sql
CREATE TABLE scans (
    id UUID PRIMARY KEY,          -- UUID generated in the code

```

### Then for inserting to the DB

```go
// Create a new scan record
	newScan := domain.NewScan()
// Do stuff like add targets, etc...

// Insert into the database
query := `INSERT INTO scans (id, name, created_at) VALUES ($1, $2, $3)`
	_, err = db.Exec(query, newScan.ID, newScan.Name, newScan.CreatedAt)
	if err != nil {
		slog.Error("Failed to insert scan", slog.Any("error", err))
	}
```

pkg/domain/scan.go
```diff
@@ -1,16 +1,14 @@
...
-	events "github.com/kptm-tools/common/common/events"
+	"github.com/kptm-tools/common/common/results"
+	"time"
 )
 
 type Metadata struct {
-	Progress string            `json:"progress"`
-	Service  enums.ServiceName `json:"service"`
+	Progress string                 `json:"progress"`
+	Service  enums.EventSubjectName `json:"service"`
```
grez-lucas, line 11: Sorry for not notifying you! I changed `enums.ServiceName` => `enums.ToolName` to use your "Tool" naming convention

pkg/domain/scan.go
```diff
@@ -21,18 +19,62 @@ type StatusHost struct {
...
 }
+
 type Scan struct {
-	ID           string          `json:"id,omitempty"`
-	HostsStatus  []StatusHost    `json:"hosts_status,omitempty"`
-	HostsResults []ResultHost    `json:"hosts_results,omitempty"`
-	Targets      []events.Target `json:"targets,omitempty"`
-	CreatedAt    time.Time       `json:"created_at"`
-	UpdatedAt    time.Time       `json:"updated_at"`
+	ID           string         `json:"id,omitempty"`
```
grez-lucas, line 24: Ideally should be uuid.UUID

pkg/domain/scan.go
```diff
@@ -21,18 +19,62 @@ type StatusHost struct {
...
+
+type SeverityCounts struct {
+	Critical int `json:"critical"`
+	High     int `json:"high"`
+	Medium   int `json:"medium"`
+	Low      int `json:"low"`
+}
+
+type ScanSummary struct {
+	ScanID          int            `json:"scan_id,omitempty"`
```
grez-lucas, line 46: uuid.UUID

pkg/domain/scan.go
```diff
@@ -21,18 +19,62 @@ type StatusHost struct {
...
+	ScanDate        string         `json:"scan_date,omitempty"`
+	Host            string         `json:"host,omitempty"`
+	Vulnerabilities int            `json:"vulnerability"`
+	Severities      SeverityCounts `json:"severity,omitempty"`
+	Duration        float64        `json:"duration,omitempty"`
+	Status          string         `json:"status,omitempty"`
+}
+
+type ScanResult struct {
+	ScanID int
```
grez-lucas, line 56: uuid.UUID

pkg/domain/scan.go
```diff
@@ -21,18 +19,62 @@ type StatusHost struct {
...
-	HostsStatus  []StatusHost    `json:"hosts_status,omitempty"`
-	HostsResults []ResultHost    `json:"hosts_results,omitempty"`
-	Targets      []events.Target `json:"targets,omitempty"`
-	CreatedAt    time.Time       `json:"created_at"`
-	UpdatedAt    time.Time       `json:"updated_at"`
+	ID           string         `json:"id,omitempty"`
+	TenantID     string         `json:"tenant_id,omitempty"`
+	OperatorID   string         `json:"operator_id,omitempty"`
+	HostID       int            `json:"host_ids,omitempty"`
+	HostsStatus  []StatusHost   `json:"hosts_status,omitempty"`
```
grez-lucas, line 28: Does `HostsStatus` and `HostResults` being slices/arrays still make sense if we're only having one host per scan?🤔

pkg/storage/scan.go
grez-lucas, line 1: Awesome work with the queries here @cristianoCatolico !

grez-lucas (approved):
Great PR @cristianoCatolico ! You made great progress on the `getScans` endpoint and even refactored most of what I was planning to do myself. There's just a couple of minor cleanup things to do before I begin incorporating event logic, such as:

- UUID for scans
- Writing sample data to test the DB queries thoroughly
- A couple of bugs regarding certain calculated fields.

I will take care of these in a separate branch this week, and merge this branch in the meantime. The comments I posted are mostly for me to reference later, don't worry about them!

PR #44, opened 2025-01-15
refactor(cmmn) Update core-service to use new EventSubject consts

grez-lucas:
These new constants and enumerations from the [latest common package tag](https://github.com/kptm-tools/common) allow us to avoid using magic strings for:

- **EventSubjects**
- **ServiceStatus** (for database scan statuses)
- **TargetTypes**
- **ServiceNames** (e.g: nmap, harvester, whois, etc.)


cristianoCatolico (approved):
Nice refactor

PR #45, opened 2025-01-22
Refactor/kt 277 scans to allow one host per scan

grez-lucas:


cristianoCatolico (approved):
Nice refactor, thanks Lucas

PR #47, opened 2025-01-24
Feature/kt 285 integrate core service event consumption

grez-lucas:
Added microservice event consumption, which means:

- Implemented interfaces for event consumption handlers
- Implemented event subscription methods using Dependency Injection pattern 
- Storing tool outputs to `scan_results`
- Storing vulnerabilities to `scan_vulnerabilities` (when the tool is an NmapEvent)

### DIFF
--- a/cmd/core-server/main.go
+++ b/cmd/core-server/main.go
@@ -6,6 +6,7 @@ import (
 	cmmn "github.com/kptm-tools/common/common/events"
 	"github.com/kptm-tools/core-service/pkg/api"
 	"github.com/kptm-tools/core-service/pkg/config"
+	"github.com/kptm-tools/core-service/pkg/events"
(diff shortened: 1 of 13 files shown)


cmd/core-server/main.go
cristianoCatolico, line 1: Nice to handle error for event

pkg/domain/scan.go
cristianoCatolico, line 1: Good helper to check if finish the scan

pkg/events/consumers/harvester.go
cristianoCatolico, line 1: Nice using slog, excelent work Lucas

pkg/storage/scan.go
cristianoCatolico, line 1: Makes sense to have success and not status since that now is in the scan table.

cristianoCatolico (approved):
Excellent work Lucas

PR #51, opened 2025-01-27
feat(scans) Add UpdateScanStatus method to service and storag

grez-lucas:
Add method to update a scan's status to `ScanService` and `ScanStorage`


cristianoCatolico (approved):
Works fine the update status

PR #52, opened 2025-01-27
Feature/kt 294 scanfailedhandler

grez-lucas:


pkg/storage/scan.go
cristianoCatolico, line 1: I did not know that we can do this 
execFunc := func(query string, args ...interface{}) (sql.Result, error) {
		if tx != nil {
			return tx.Exec(query, args...)
		} else {
			return s.db.Exec(query, args...)
		}
	}
I am surprised, nice coding

cristianoCatolico (approved):
As usual, good refactor and implementation of handler for scan failed

pkg/storage/scan.go
grez-lucas: It would be awesome to extract it into some abstract method so that we can use transactions conditionally!

PR #53, opened 2025-01-28
Feature/kt 247 cancel scan endpoint

grez-lucas:


cristianoCatolico (approved):
Good work Lucas, so now is just to finish the trigger

PR #54, opened 2025-01-29
[feat]: verification of status scan

cristianoCatolico:
This is the script for SQL for updating status. 

```sql
CREATE OR REPLACE FUNCTION scan_status_changes()
  RETURNS TRIGGER
  LANGUAGE PLPGSQL
  AS
$$
        DECLARE quantityTools integer;
        DECLARE quantityScanResultPerTool integer;
BEGIN
    select count(*) from tools into quantityTools;
    select count(*) from scan_results where scan_id= NEW.scan_id into quantityScanResultPerTool;
    IF quantityScanResultPerTool = 0 THEN
        update scans set status='InProgress' where id = NEW.scan_id;
    END IF;
    IF quantityTools = quantityScanResultPerTool THEN
        update scans set status='Completed' where id = NEW.scan_id;
    END IF;
RETURN NEW;
END;
$$
```

### DIFF
--- a/pkg/storage/scan.go
+++ b/pkg/storage/scan.go
@@ -554,7 +554,6 @@ func (s *PostgreSQLStore) UpdateScanStatusAndEndedAt(tx *sql.Tx, scanID uuid.UUI
 	return nil
 }
 
-<<<<<<< HEAD
 func (s *PostgreSQLStore) GetScanInsights(scanID uuid.UUID) (*domain.ScanInsights, error) {
 	query := `
     WITH severity_per_type AS (
--- a/pkg/storage/scan.go
+++ b/pkg/storage/scan.go
@@ -554,6 +554,7 @@ func (s *PostgreSQLStore) UpdateScanStatusAndEndedAt(tx *sql.Tx, scanID uuid.UUI
 	return nil
 }
 
+<<<<<<< HEAD
 func (s *PostgreSQLStore) GetScanInsights(scanID uuid.UUID) (*domain.ScanInsights, error) {
 	query := `
     WITH severity_per_type AS (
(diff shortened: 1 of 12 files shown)


grez-lucas (changes requested):
Quite simple and good check of the status before insertion! There are a couple of things that can be improved:

## Handlers Updates

This block of code in EventHandler functions seems to repeat a lot, without any particular logic being different. 
```go
	statusNotToUpdateResult := []string{enums.StatusFailed.String(), enums.StatusCancelled.String()}
		if slices.Contains(statusNotToUpdateResult, actualScan.Status) {
			slog.Error("Error inserting ScanResult to DB because of Scan Status",
				slog.String("scan_id", evt.ScanID.String()),
				slog.String("tool_name", string(evt.ToolResult.Tool)),
				slog.Any("Status ", actualScan.Status),
			)
```

While we may come across a case in the future where a particular tool may bypass this check, I suggest you extract this logic into a single function within the `events` package which can be later on called from these handlers. This will:

1. Make it easier to maintain code, contrary to having to change the same logic in 4 different files
2. Keep our code a little cleaner

Something like:

pkg/events/events.go
```go
func CanInsertScanResult(currentStatus string) bool {
 statusNotToUpdateResult := []string{enums.StatusFailed.String(), enums.StatusCancelled.String()}
if !slices.Contains(statusNotToUpdateResult, actualScan.Status) {
return true
}
return false
}
```

Then you can call that from each handler and handle/log accordingly. 


## SQL Trigger

I don't see any piece of code that adds your trigger programatically to the database, which should be done when the tables are set up. This will help me quality-test and make sure that the trigger is actually implemented in other dev environments. Here's a little help on how to do it ⭐


```go
	// SQL query to create the trigger
	createTriggerQuery := `
	CREATE OR REPLACE FUNCTION my_trigger_function()
	RETURNS TRIGGER AS $$
	BEGIN
		-- Example logic: Log the insert event
		INSERT INTO my_log_table (action, table_name, created_at)
		VALUES ('INSERT', TG_TABLE_NAME, NOW());
		RETURN NEW;
	END;
	$$ LANGUAGE plpgsql;

	CREATE TRIGGER my_table_insert_trigger
	AFTER INSERT ON my_table
	FOR EACH ROW
	EXECUTE FUNCTION my_trigger_function();
	`

	// Execute the query to create the trigger
	_, err = db.Exec(createTriggerQuery)
	if err != nil {
		log.Fatal("Error creating trigger: ", err)
	}

	fmt.Println("Trigger created successfully.")
```

Aside from that, please make sure your trigger should handle the following edge cases:

#### 1. Trigger Runs on Every Insert, Even If Status is Already 'Completed'
- Fix: Add a condition to check the currentStatus and skip updates if the scan is already 'Completed'
#### 2. Incorrect Status Update When First Result Arrives
- The trigger sets the scan status to 'InProgress' **only if** there are zero results (quantityScanResultPerTool = 0).
- This means that when the first result is inserted, the status won't be updated unless the scan was originally empty.

Remember that in SQL:
- When you insert a new row into the scan_results table, the trigger will fire after the insert operation completes.


Possible fix:

```sql
CREATE OR REPLACE FUNCTION scan_status_changes()
RETURNS TRIGGER
LANGUAGE PLPGSQL
AS
$$
DECLARE
    quantityTools INT;
    quantityScanResults INT;
    currentStatus TEXT;
BEGIN
    -- Get the total number of tools
    SELECT COUNT(*) INTO quantityTools FROM tools;
    
    -- Get the number of scan results for this scan
    SELECT COUNT(*) INTO quantityScanResults FROM scan_results WHERE scan_id = NEW.scan_id;
    
    -- Get current scan status
    SELECT status INTO currentStatus FROM scans WHERE id = NEW.scan_id;

    -- If scan is already completed, do nothing
    IF currentStatus = 'Completed' THEN
        RETURN NEW;
    END IF;

    -- Update status to 'InProgress' if at least one result exists and it's not yet 'InProgress'
    IF quantityScanResults > 0 AND currentStatus NOT IN ('InProgress', 'Completed') THEN
        UPDATE scans SET status = 'InProgress' WHERE id = NEW.scan_id;
    END IF;

    -- If all expected results are present, mark as 'Completed'
    IF quantityScanResults >= quantityTools THEN
        UPDATE scans SET status = 'Completed' WHERE id = NEW.scan_id;
    END IF;

    RETURN NEW;
END;
$$;
```

cristianoCatolico (commented):
Thanks for your observations Lucas

pkg/utils/events/utils.go
grez-lucas, line 1: Personally I've always thought that having a `/utils` directory is kind of messy, since in practice developers just end up throwing all shared code there 🤣. I would have preferred for this to be defined in the `pkg/events/events.go` package or in `pkg/events/consumers.go` package since only consumers  use this piece of code. This is an improvement anyway

pkg/storage/storage.go
grez-lucas, line 1: Nice! It seems our "Create" logic in our repository pattern (storage) has grown quite a bit. I'll look into refactoring this into migrations for ease of use and version control by integrating the [golang-migrate package](https://pkg.go.dev/github.com/golang-migrate/migrate/v4) later this or next week

grez-lucas, line 1: Nice! It seems our "Create" logic in our repository pattern (storage) has grown quite a bit. I'll look into refactoring this into migrations for ease of use and version control by integrating the [golang-migrate package](https://pkg.go.dev/github.com/golang-migrate/migrate/v4) later this or next week

grez-lucas, line 1: Nice! It seems our "Create" logic in our repository pattern (storage) has grown quite a bit. I'll look into refactoring this into migrations for ease of use and version control by integrating the [golang-migrate package](https://pkg.go.dev/github.com/golang-migrate/migrate/v4) later this or next week

pkg/storage/scan.go
grez-lucas, line 1: The `ended_at` column should be updated to the current timestamp when the status becomes 'Completed'

grez-lucas (changes requested):

grez-lucas (approved):

PR #55, opened 2025-01-29
Feature/kt 295 scan summary endpoint

grez-lucas:


Dockerfile
cristianoCatolico, line 1: we rise version, wow time passes quickly

cristianoCatolico (approved):
Very big feature, nice implementation of get scan insights.

PR #68, opened 2025-02-04
Refactor/kt 318 db migrations with golang migrate package

grez-lucas:
## Main changes 📂🔧

* Moved database table/trigger creation logic to migrations, which allows us to improve maintainability and scalability.
* Added golang-migrate logic to `main.go`, which embeds the `/migrations` directory into our code and reads from them so that there's no need to copy directories in a production environment.
* Added Makefile helper commands to help run and create migrations.
* Minor fixes to `MarkAsCancelled` and `MarkAsFailed` to update ended_at time.Time with `UTC()`
* Refactored environment variable management to avoid bugs in production regarding default values and to enforce more secure practices

## Database changes 🔄🗄️
- Replaced `Tools` table with a Tool enum, which is more efficient since we're not doing anything with a tool's description or type
- Replaced `scan_status` column from VARCHAR to an enum which follows the definition in `common`
- Replaced `port_status` with an enum 

The app runs migrations automatically on startup, but if you wish to run them manually via the `Makefile` commands, you must install [The Migrate CLI](https://github.com/golang-migrate/migrate)


cristianoCatolico (approved):
Great refactor using migrations, for tools using types, I did not know that postgres has that, very interesting refactor. Congratulations for this great work.

PR #71, opened 2025-02-07
Refactor/cleanup host validation

grez-lucas:
- Extracted validation logic to common package. Core-service and other microservices call the `HostClassification` methods to validate the host properly now.
- Refactored common module, which meant changing some import statements throughout the codebase.
- Moved some methods from `hostsHandler` to `hostService` such as fetching the Hostname and IP of the host, since the service should be in charge of business logic, not the handler.


cristianoCatolico (approved):
Interesting, I have not noticed that var with := is not necessary.  Also the use of config never notice of calling many times. The part of validation of host, very good enhancement.

PR #75, opened 2025-02-13
Feature/kt 319 implement reports endpoint

grez-lucas:
## ✨ Features
- Implemented `api/reports` endpoint, which uses scan handlers and scan service.
- Added domain structs for `ReportItem` (for business logic)
- Added DTO struct for `ReportsReponse` (for API resources)
- Implemented storage query to get reports for a given TenantID and parse a vulnerability's comment status.
## 💾 Storage
- Added a simple `TEXT` column to the `scan_vulnerabilities` table called `analyst_comment` for an analyst to post a comment for a certain scan's vulnerability. By default it is `NULL`.
  -  **Reasoning:** Requiements don't currently necessitate anything too complex for comments at the moment (such as comment date, or comment user). This can later be refactored into it's own `scan_vulnerability_comments` table layer if requirements change.


cristianoCatolico (approved):
Excellent job, now we have the part of comments and one report done, and the test for scan service. Thanks Lucas!

PR #76, opened 2025-02-14
Feature/kt 322 add scorecard trends endpoint

grez-lucas:
## ✨ Features
- Added an endpoint to get the scorecard trends of all hosts associated to a tenant
- Added functions to get the oldest and latest scans associated to a hostID, with conditional fromDate and toDate logic
- Added unit tests for handler parsing of http request parameters
## ♻️ Refactors
- Added new samples to the samples package
- Updated samples so that results are consistent with hostnames and IP's


cristianoCatolico (approved):
I trust that calculate of score is good , the  code as always is clean.

PR #78, opened 2025-02-14
Feature/kt 314

cristianoCatolico:
This feature contains scheduling with repetition and not repetition.


pkg/storage/listener.go
```diff
@@ -59,21 +81,77 @@ func (pl *PostgresListener) startListening() {
...
+			}
 
-		slog.Debug("Parsed scan completed event", slog.String("scanID", scanCompletedEvent.ScanID.String()))
-		// Use scanService to handle scanCompleted
-		if err := pl.scanService.HandleScanCompletion(scanCompletedEvent.ScanID); err != nil {
-			slog.Error("Failed to handle scan completion",
-				slog.String("scanID", scanCompletedEvent.ScanID.String()),
-				slog.Any("error", err))
+			slog.Debug("Parsed scan completed event", slog.String("scanID", scanCron.ScanID.String()))
+			// Use scanService to handle scanCompleted
```
grez-lucas: The business logic (getting the Host by ID, and then creating a target should be a call to a scanService method, which calls to these storage methods. We should keep this business logic inside our service layer, and avoid coupling our listener to the storage layer.

Ideally, only the service should be coupled with the storage layer.

cmd/main.go
grez-lucas, line 1: Good stuff using dependency injection here! Overall, I feel the coreStore should not be injected though, because it is already being injected as a dependency into `scanService`. 

Instead, it would be cleaner to do the following:
```go
storage.NewPostgresListener(c, scanService, eventBus)
```
* Now the listener doesn't depend on both the `scanService` and the `coreStore`
   * In order for the listener to run our **business logic** (like getting a host by ID, and then creating a target from that host), we should write a `scanService` method, which is called from the listener. The listener should not call our storage layer directly.

grez-lucas: This also makes unit testing our `scanService` easier ;)

cmd/migrations/000012_create_scan_scheduling_table.down.sql
grez-lucas, line 1: Please move these migrations into the `cmd/migrations/migrations` directory and rename their file id's accordingly.

pkg/handlers/requests.go
grez-lucas, line 1: This request struct is failing to take two considerations into account:

1. We can have multiple `HostIds`, but how do we know which `ScheduleAt` and `IsRepeat` is mapped to each one of the hostID's?
2. ScheduleAt and IsRepeat are optional values, which means `decodeJSONBody` will throw a BadRequest if these are missing.

I see two options here:

1. **Refactor this endpoint to create ONE scan**: And update documentation for @diegocarhuavilca to call upon it several times, one for each host.
2. Create a ScanRequest object which is associated to only one host, and then make this struct have several of those (with optional parameters):

e.g:
```go
type ScanRequest struct {
    HostID   int `json:"host_id"` // hostID is an int
    ScheduleAt *string  `json:"schedule_at"` // Pointer to string
    IsRepeat   *bool    `json:"is_repeat"`   // Pointer to bool
}
```

Then check if `ScheduleAt==nil` for your conditional logic.

Honestly I would go with 1. since this endpoint is now a little more complex (and also keep the pointer/optional values). But any option is fine.

pkg/interfaces/scan.go
grez-lucas, line 1: The method to Run a scan/get a host's target (which I mentioned above), should go in this interface and be called from the listener.

pkg/storage/listener.go
```diff
@@ -40,12 +57,17 @@ func NewPostgresListener(
...
 	slog.Info("PostgresListener started", slog.String("channel", "scan_completed"))
 
+	if err := listener.Listen("scan_cron"); err != nil {
+		return nil, fmt.Errorf("failed to listen to scan_cron channel: %w", err)
+	}
+	slog.Info("PostgresListener started", slog.String("channel", "scan_cron"))
 	postgresListener := &PostgresListener{
 		listener:    listener,
 		scanService: scanService,
+		storage:     storage,
```
grez-lucas: As mentioned before, it would be cleaner to avoid injecting the storage into this listener

pkg/storage/scan.go
grez-lucas, line 1: Looks pretty good!

grez-lucas (changes requested):
Very good first approach for this task! 

There are some key software design principles which can be improved to improve modularity, testability and portability. 
There are also some errors which must be taken into account, such as the correct location for your migration files. 

Please fix these and [Update the Apiary Documentation](https://kptmtools.docs.apiary.io/) before requesting a new review ⭐

cmd/migrations/000012_create_scan_scheduling_table.down.sql
cristianoCatolico: Done

pkg/storage/listener.go
cristianoCatolico: Done

pkg/storage/listener.go
cristianoCatolico: Done

cmd/main.go
cristianoCatolico: Done

pkg/handlers/requests.go
cristianoCatolico: I done the first one, also I update the documentation

pkg/storage/listener.go
```diff
@@ -3,23 +3,38 @@ package storage
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/google/uuid"
+	"github.com/kptm-tools/common/common/pkg/enums"
```
grez-lucas, line 7: This package is being imported more than once

pkg/handlers/requests.go
```diff
@@ -64,5 +64,7 @@ type ServiceHost struct {
 	Host  string   `json:"host"`
 }
 type ScanRequest struct {
-	HostIds []string `json:"host_ids"`
+	HostId     int     `json:"host_id"`
```
grez-lucas: struct field `HostId` should be `HostID` [(ST1003)](https://staticcheck.dev/docs/checks/#ST1003)

Please don't forget we have an integrated audit static and vulnerability checker in the app, which you can run using `make audit` to avoid these kinds of errors.

```suggestion
	HostID     int     `json:"host_id"`
```

pkg/handlers/scan.go
grez-lucas, line 1: Good job with this handler! Things are looking cleaner now. At the handler level, we always want to validate that the request parameters are in the right format. Our `decodeJSONBody` helper function makes sure that struct fields are present, but there is one validation missing here.

Since   we have the field:
```go
  ScheduleAt *string  `json:"schedule_at"` // Pointer to string 
```
We should validate before passing to service that this time string is in the right format. You can do this [using the `time.Parse()` method](https://pkg.go.dev/time#Parse) from the `time` package. I have a small example of this for validating date ranges [here](https://github.com/kptm-tools/core-service/blob/3c2490921e4948fc0f0b7472fef8e2fef986e5ea/pkg/handlers/scan.go#L276). 
- The endpoint should consider **and document** a [time layout](https://pkg.go.dev/time#Layout), which must include date and time. (Perhaps `DateTime` could be a good choice?). Right now there are no clues as to what string format you're asking for.
- Don't forget to conditionally validate, since `ScheduleAt` could be **nil** and this could lead to our program exploding!

pkg/services/scan_service_test.go
grez-lucas, line 1: It would be nice to see you create some unit tests for scanService.CreateScan ;)

pkg/handlers/scan.go
grez-lucas: With this in mind, these `interfaces/scan` and `interfaces/storage` modifications would make sense:

#### `InsertScanScheduling(scans *domain.Scan, scheduleAt string, isRepeated bool) error`
- ✅`InsertScanScheduling(scans *domain.Scan, scheduleAt time.Time, isRepeated bool) error`

#### `CreateScanScheduling(uuid.UUID, string, bool) error`
- ✅`CreateScanScheduling(uuid.UUID, string, bool) error` -> here you would have your logic to pass `time.Time` to cronTime or whatever

grez-lucas: Also, both the `ScheduleAt` and `IsRepeat` values should have nil checks before the call to the service!

pkg/handlers/scan.go
```diff
@@ -68,23 +57,31 @@ func (h *ScanHandlers) CreateScans(w http.ResponseWriter, req *http.Request) err
...
 		}
 		scanStartedBytes, err := json.Marshal(scanStartedPayload)
 		if err != nil {
 			return api.WriteJSON(w, http.StatusInternalServerError, err.Error())
 		}
 		h.eventBus.Publish(string(enums.ScanStartedEventSubject), scanStartedBytes)
 
+	} else {
+		errScanSchedule := h.scanService.InsertScanScheduling(scan, *scanRequest.ScheduleAt, *scanRequest.IsRepeat)
+		if errScanSchedule != nil {
```
grez-lucas, line 116: I think there may be other kinds of errors which may come from the service, other than just invalid scheduling, we should make this error handling a bit more generic, returning a `BadRequest` response if the scheduling is invalid beforehand (as mentioned in the parent comment), and handling InternalServer errors with their message here. (Also logging!)

```suggestion
		if scanRequest.ScheduleAt == nil && scanRequest.IsRepeat == nil { // Or a better conditional check for both attributes

		scanStartedPayload := &cmmn.ScanStartedEvent{
			BaseEvent: cmmn.BaseEvent{
				ScanID:    scan.ID,
				Timestamp: scan.CreatedAt.UTC(),
			},
			Target: scan.Target,
		}
		scanStartedBytes, err := json.Marshal(scanStartedPayload)
		if err != nil {
			return api.WriteJSON(w, http.StatusInternalServerError, err.Error())
		}
		h.eventBus.Publish(string(enums.ScanStartedEventSubject), scanStartedBytes)

	} else {
		err := h.scanService.InsertScanScheduling(scan, *scanRequest.ScheduleAt, *scanRequest.IsRepeat)
		if err != nil {
Comment
			slog.Error("Error inserting ScanScheduling",
			                      slog.Int("host_id", scanRequest.HostID),
			                      slog.Any("error", err))
			return api.WriteJSON(w, http.InternalServerError, api.APIError{Error: http.StatusText(http.InternalServerError)})
		}
	}
```

grez-lucas: Make sure too keep in mind cases where:
- `ScheduleAt == nil` but `IsRepeat != nil` and viceversa

pkg/storage/scan.go
grez-lucas, line 1: `obtainYearCron` definetely **needs units testing** if you plan on integrating it. Working with strings might be messy, I suggest you refactor this to `func obtainYearCron(scheduledAt time.Time) (int, string, error)` to make things clearer (I had to decipher what string format you were asking for). 

There are a couple of packages I found which might have helped you with a different approach, but at this point it might require a bigger refactor.
- [gocron](https://github.com/go-co-op/gocron)

grez-lucas (changes requested):
Looks better. There are some handler validation checks which can be improved and pure functions such as `obtainYearCron` which **should have unit tests** done. 

Please fix these suggestions and improve documentation with examples, since some fields such as `ScheduleAt *string` don't have any documentation written on the kind of format they expect, as well as the fact that they can be **nullable**.

pkg/storage/listener.go
cristianoCatolico: Done

pkg/handlers/requests.go
cristianoCatolico: I forgot to use that audit,  I will do it before sending to you

pkg/domain/scan.go
grez-lucas, line 1: This is great, though I feel it would benefit from being left in a `/domain/scanschedule.go` file

pkg/handlers/requests.go
grez-lucas, line 1: I think this change hasn't been documented in our swagger docs.

pkg/services/scan_schedule_service_test.go
grez-lucas, line 1: These unit tests are shorter than I expected 😂

grez-lucas (commented):
Looks good overall and quite ready to merge, we just need some unit tests for our domain logic found in `pkg/services/scanschedule.go`. 

Both `pkg/services/vulnerability_test.go` and `pkg/services/scan_service_test.go` should give you a good idea on how to use the mock storage for unit testing for your `pkg/services/scanschedule.go` methods. Please consider success, error and edge cases in your unit tests. With that done we should be good to merge!

pkg/handlers/scanschedule_test.go
grez-lucas, line 1: Good work using `httptest` and our mocks to test out this handlers! ;)

grez-lucas, line 1: Good work using `httptest` and our mocks to test out this handlers! ;)

pkg/handlers/scanschedule_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/kptm-tools/core-service/pkg/mocks"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestScanScheduleHandlers_Valid(t *testing.T) {
```
grez-lucas, line 16: Nice unit test! The name is a little bit misleading, since it includes both valid and Invalid Dates

pkg/mocks/scanschedule_service.go
grez-lucas, line 1: Nice

pkg/services/scan_schedule_service_test.go
grez-lucas, line 1: Good work with these unit tests! Quite thorough :)

grez-lucas (approved):
Very good work @cristianoCatolico ! I'm very proud that you managed to add dependency-injection based unit tests to both the new service and handler. I hope this becomes a standard going forward for future features we add ;)

There's only some minor details regarding the docs: 

image


Where I believe the "repeat_frequency" object is not properly mentioned, since the request for scheduling a scan looks something more like:

```json
{
    "host_id": 1,
    "schedule_at": "2025-03-05T18:04:59.000Z",
    "repeat_frequency": {
        "quantity": 3,
        "unit_of_frequency": "day"
    }
}
```

Please document this, and notify @diegocarhuavilca so he can get to work refactoring and integrating Scan-creation-related views!

PR #90, opened 2025-03-06
feat: test of email service

cristianoCatolico:
This pr contains service to send emails and it is being used in cancel and complete status of an san.


pkg/config/config.go
grez-lucas, line 1: We should use `fetchEnvOrPanic` for SMTP credentials. 

Even though in a **Overall System Context** it makes sense that these credentials are part of the FusionAuth config, in our `core-service` context I think they should go in a different part of the config struct. Something like `SMTP`, with subfields for Host, Port, User, etc. Right now they're part of the FusionAuth substruct. Please add a new sub-struct (like I did with Database, Nats, FusionAuth, but specifically for SMTP.

pkg/mocks/email_service.go
grez-lucas, line 1: 👍🏻

pkg/storage/scan.go
```diff
@@ -1091,3 +1091,18 @@ func (s *PostgreSQLStore) UpdateScanScheduling(scanID uuid.UUID, scanScheduleID
...
 }
+
+func (s *PostgreSQLStore) GetRapporteursScan(scanID uuid.UUID) (*[]domain.Rapporteur, string) {
+	query := `SELECT H.rapporteurs, H.alias FROM scans S INNER JOIN  hosts H ON H.id = S.host_id WHERE S.id=$1 `
+	var rapporteursBytes []byte
+	var rapporteurs []domain.Rapporteur
+	var name string
+	err := s.db.QueryRow(query, scanID).Scan(&rapporteursBytes, &name)
+	if err != nil {
+		return nil, ""
```
grez-lucas: Our app should handle these errors!

pkg/interfaces/storage.go
grez-lucas, line 1: It would be a bit more idiomatic (but verbose) for the signature to be:

`GetRapporteursAndHostAliasByScanID(scanID uuid.UUID) ([]*domain.Rapporteur, string, error)`

pkg/storage/scan.go
grez-lucas, line 1: Error handling could be improved in this storage method.  Consider the previous interface function signature I mentioned. Please take these errors into account:

1. **No rows are found:** `s.db.QueryRow` will return an `sql.ErrNoRows`, in which case the host was not found for that scan. Here we should return a `customerrors.ErrHostNotFound` and handle that accordingly in the called function. 
2. **Query error:** Here we should return no results, and notify the called (with the new function signature) by wrapping the error `fmt.Errorf()`
3. **json.Unmarshal error:** In case data is corrupted, or simply un-unmarshallable, we should also notify the caller that this method failed.

pkg/interfaces/storage.go
grez-lucas: Using `[]*domain.Rapporteur` instead of `*[]domain.Rapporteur` should help with weird errors were the slice is nil for some reason.

pkg/handlers/scan.go
```diff
@@ -170,6 +173,13 @@ func (h *ScanHandlers) CancelScanByID(w http.ResponseWriter, req *http.Request)
 		}
 		return api.WriteJSON(w, http.StatusInternalServerError, api.APIError{Error: err.Error()})
 	}
+	// 3. Get the emails rapporteurs structure
+	rapporteurs, hostName := h.scanService.GetRapporteursScan(scanID)
+	// 4. Notify rapporteurs of host associated
+	errSendEmail := h.emailService.SendEmail(rapporteurs, fmt.Sprintf("Scan %s has been cancelled", scanID), fmt.Sprintf("Dear Rapporteur, your scan associated to host %s has been cancelled", hostName))
+	if errSendEmail != nil {
+		slog.Error("Problems with sending email", slog.Any("error", errSendEmail.Error()))
```
grez-lucas: ScanID and Rapporteurs would be useful to add to this log too

pkg/handlers/scan.go
grez-lucas, line 1: I like this approach of using the `SendEmail` method, but it does make the code for this handler longer and a bit less readable. I feel having separate methods for `SendScanFailedEmail`, `SendScanCompleteEmail` or `SendScanCancelledEmail` would be cleaner, that way the **handler doesn't have to worry about the body and subject of the email.**

pkg/interfaces/email.go
grez-lucas, line 1: Quoting the previous comment:

> I like this approach of using the SendEmail method, but it does make the code for this handler longer and a bit less readable. I feel having separate methods for SendScanFailedEmail, SendScanCompleteEmail or SendScanCancelledEmail would be cleaner, that way the handler doesn't have to worry about the body and subject of the email.

You shoud keep this `SendEmailMethod` internally in the `pkg/services/email.go` file, but keep the logic for email contents inside of other exposed functions, such as: 
- `SendScanFailedEmail(string recipient) error`
-  `SendScanCompleteEmail(string recipient) error`
- `SendScanCancelledEmail(string recipient) error`

Then **you can expose these functions via our interface**, and keep the logic for each for each subject and body inside, which will make refactoring (if email contents change later) easier. Internally, each of these functions would call `SendEmail()`, but whoever calls these functions from our interface has to worry less about internal logic ;)

grez-lucas: Also, since the signature for each method is just **one recipient**, we would just have to loop over the Rapporteur emails in the caller function. This could make logging a bit easier, letting us know if a particular rapporteur's email had trouble, without making the app explode


In caller:
```go
for _, email := range rapporteurEmails {
   if err := SendScanCompletedEmail(email); err != nil {
      slog.Warn("Failed to send email to rapporteur", 
      slog.String("scan_id", scan.ID),
      slog.String("rapporteur_email", email),
      slog.Any("error", err),
       )
       continue
       // Rest of the code
}
}

File: pkg/services/email.go [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+	addresses := make([]string, sizeAddress)
+	for i, recipient := range *toAddress {
+		addresses[i] = m.FormatAddress(recipient.Email, recipient.Name)
+	}
+	m.SetHeader("From", s.FromEmail)
+	m.SetHeader("To", addresses...)
+	m.SetHeader("Subject", subject)
+	m.SetBody("text/html", body)
+
+	if err := s.SendMail(m); err != nil {
```
grez-lucas, line 70: It would be more idiomatic to return this error and let the caller handle logging. Right now the caller only knows if there was an error if there's no toAddresses, but doesn't get to know if there's a network, authentication, etc error sending the email.

pkg/services/email.go
grez-lucas, line 1: This service's main method `SendEmail` should consider a couple of errors (in particular if the library you're using includes specific errors):

1. **Transient Errors (Retriable):** These are temporary issues that might resolve themselves if you retry the email sending. Examples: network glitches, temporary service unavailability, rate limits. In this case, we should have a simple retry mechanism.

2. **Permanent Errors (Non-Retriable):** These are errors that indicate a fundamental problem that retrying won't fix. Examples: authentication failures, invalid recipient addresses, server rejects. In this case, we must return the error.

3. **Invalid Recipient Errors**

grez-lucas (changes requested):
Solid first approach! There's some improvements to be made in order to make code more **idiomatic and modular**. Please address feedback before submitting another request 😁

pkg/services/email.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	} else {
+		slog.Info("Email sent")
+	}
+	return nil
+}
+
+func (s *EmailService) SendScanFailedEmail(recipient string, hostName string) error {
+	subject := "Scan Failed of" + hostName
+	body := "Dear Recipient, the scan has been failed"
```
grez-lucas:
```suggestion
	body := "Dear Recipient, the scan has failed"
```

pkg/services/email.go
grez-lucas, line 1: Good work on using custom errors :)

pkg/services/email_test.go
grez-lucas, line 1: These tests are enough, but some testing for the retry mechanism and custom error returns would've been nice to see.

pkg/services/email.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	}
+	slog.Info("Email of Scan failed sent", slog.String("recipient", recipient))
+	return nil
+}
+
+func (s *EmailService) SendScanCompletedEmail(recipient string, hostName string) error {
+	subject := "Scan Completed of " + hostName
+	body := "Dear Recipient, the scan has been completed"
+	recipientRapporteur := []*domain.Rapporteur{
```
grez-lucas: These struct literal is using unkeyed fields. While this is not a compile time error, it does have bad code smell and reduces readability, please use keys when defining struct literals:
```suggestion
	recipientRapporteur := []*domain.Rapporteur{
	  Name: "",
	  Email: recipient,
	  IsPrincipal: false,
	 }
```

pkg/services/email.go
```diff
@@ -0,0 +1,175 @@
...
+	slog.Info("Email of Scan completed sent", slog.String("recipient", recipient))
+	return nil
+}
+
+func (s *EmailService) SendScanCancelledEmail(recipient string, hostName string) error {
+	subject := "Scan cancelled of " + hostName
+	body := "Dear Recipient, the scan has been canceled"
+	recipientRapporteur := []*domain.Rapporteur{
+		{
+			"",
```
grez-lucas: Same here

pkg/services/email.go
```diff
@@ -0,0 +1,175 @@
...
+		slog.Info("Email sent")
+	}
+	return nil
+}
+
+func (s *EmailService) SendScanFailedEmail(recipient string, hostName string) error {
+	subject := "Scan Failed of" + hostName
+	body := "Dear Recipient, the scan has been failed"
+	recipientRapporteur := []*domain.Rapporteur{
+		{
```
grez-lucas: Same as comment below ⬇️

pkg/services/email_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	emailService := &EmailService{
+		Host:     "smtp.example.com",
+		Port:     "587",
+		Username: "your-email@example.com",
+		Password: "your-email-password",
+		SendMail: mockSMTPClient.SendMail,
+	}
+
+	err := emailService.SendEmail([]*domain.Rapporteur{
```
grez-lucas: These struct literal is using unkeyed fields. While this is not a compile time error, it does have bad code smell and reduces readability, please use keys when defining struct literals:
```suggestion
	rapporteurs := []*domain.Rapporteur{
	  Name: "",
	  Email: recipient,
	  IsPrincipal: false,
	 }
```

pkg/services/email.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	m.SetHeader("From", s.FromEmail)
+	m.SetHeader("To", addresses...)
+	m.SetHeader("Subject", subject)
+	m.SetBody("text/html", body)
+
+	if err := s.SendMail(m); err != nil {
+		return err
+	} else {
+		slog.Info("Email sent")
```
grez-lucas: This log could be more informative

pkg/services/email.go
```diff
@@ -0,0 +1,175 @@
...
+		Host:      host,
+		Port:      port,
+		Username:  username,
+		Password:  password,
+		FromEmail: fromEmail,
+		SendMail:  dialer.DialAndSend,
+	}
+}
+
+func (s *EmailService) SendEmail(toAddress []*domain.Rapporteur, subject, body string) error {
```
grez-lucas: As stated in the previous PR review, `SendEmail` currently sends many emails to many addresses. There are spelling mistakes which should be fixed to improve readability if this behavior is intended:

1. `SendEmail(toAddress ....)` -> `SendEmails(toAddresses .....)`

Again, this handles many emails being sent, but returns an error immediately on the first s.SendEmail(m) error. This is bad code smell, think about this case:

I have the following rapporteur addresses: [ `validemail@example.com`, `invalidemail@example.com`, `validemail2@example.com` ]. 

Because your code loops through each email and returns immediately if  there's an error, it would stop on `invalidemail@example.com` and never try to send an email to `validemail2@example.com`. 

## Suggestion

My suggestion is that you refactor this method and **remove unnecessary multiple address logic**, **only focus on sending ONE email to ONE rapporteur**. Leave multiple address logic for this method's caller. This will:
- Make your code more modular
- Make error handling easier
- Improve code readability

The new signature for this method would be something like `func (s *EmailService) SendEmail(toRapporteur domain.Rapporteur, subject, body string) error`

Notice how:
- We don't use a pointer, since we're not modifying data, just reading
- We're not using slices, we just handle the case for one rapporteur (leaving multiple rapporteur responsibility to the caller)
- The function's parameters now better reflect what we're passing to the function

pkg/services/email.go
```diff
@@ -0,0 +1,175 @@
...
+	if toAddress == nil || len(toAddress) == 0 {
```
grez-lucas: Should omit nil check; len() for nil slices is defined as zero

grez-lucas: Also unnecessary if we choose to go with modular approach ;)

pkg/services/email.go
grez-lucas: Of course, this works right now because you're _already_ looping through rapporteurs at the handler level! (see [pkg/handlers/scan.go](https://github.com/kptm-tools/core-service/pull/90/files#diff-57b8bbc73c5aa889cb7d57bf22e31473543333f2a543542009aca7fbc8f3499e)). But it seems you forgot to update this function's signature and logic.

grez-lucas (changes requested):
This looks much better. Please address the following minor issues:

1. [GitGuardian Security Checks](https://github.com/kptm-tools/core-service/pull/90/checks?check_run_id=38579266147) 😬
2. `services/email.go` Improvements on `SendEmail()` modularity and audit static check errors.
3. English spelling mistakes on email contents.

PR #92, opened 2025-03-12
feat: test of email service without secret

cristianoCatolico:


grez-lucas (approved):
Good improvements! Thanks for taking code suggestions into consideration.

PR #94, opened 2025-03-13
Bugfix/kt 375 vuln summary trends average should be based on existing data

grez-lucas:
## PR Summary: Differentiate 0 vs NULL Vulnerability Counts in Trends

**Backend Changes:**

This PR modifies the vulnerability trends query to accurately distinguish between:

*   **0 Vulnerabilities:** A scan occurred within the time period and found no vulnerabilities.
*   **NULL Vulnerabilities:** No scan was performed within the time period.

**Key Changes in Query:**

*   **Introduced `ScanPeriods` CTE:** This new CTE identifies time periods with scan data, ensuring we only consider periods where scans were actually performed when counting vulnerabilities.
*   **Modified `VulnerabilityCounts` CTE:**  This CTE now joins with `ScanPeriods` and counts vulnerabilities based on scans within those identified periods.
*   **Removed `COALESCE` in final `SELECT`:** By removing `COALESCE`, the query now returns `NULL` for `vulnerability_count` when there is no corresponding scan in a time period, allowing the frontend to differentiate "no data" from a zero count.

**Impact:**

This enhancement provides more accurate vulnerability trend data, allowing users to distinguish between periods with no vulnerabilities and periods without scan data, improving clarity and insights in vulnerability reporting.


cristianoCatolico (approved):
Very complex query, and nice enhancement for the query.

PR #95, opened 2025-03-13
feat: crud of comment vulnerability

cristianoCatolico:


pkg/handlers/vulnerability.go
```diff
@@ -82,3 +83,88 @@ func (h *VulnerabilityHandlers) GetVulnerability(w http.ResponseWriter, r *http.
...
+		}
+	}
+	isCreated, errCreate := h.vulnerabilityService.CreateVulnerabilityComment(vulnID, vulnCommentRequest.Comment)
+	if errCreate != nil {
+		if errors.Is(errCreate, sql.ErrNoRows) {
+			statusCode := http.StatusNotFound
+			return api.WriteJSON(w, statusCode, api.APIError{Error: http.StatusText(statusCode)})
+		}
+		slog.Error("Failed to create vulnerability comment ", slog.Any("error", errCreate))
+		return api.WriteJSON(w, http.StatusInternalServerError, errCreate.Error())
```
grez-lucas: Should use `APIError` struct to keep response body standard

pkg/handlers/vulnerability.go
```diff
@@ -82,3 +83,88 @@ func (h *VulnerabilityHandlers) GetVulnerability(w http.ResponseWriter, r *http.
...
+	}
+	result := make(map[string]bool)
+	result["created"] = isCreated
+	return api.WriteJSON(w, http.StatusOK, result)
```
grez-lucas: Could've used `http.StatusText(http.StatusCodeOK)` for body

pkg/handlers/vulnerability.go
```diff
@@ -82,3 +83,88 @@ func (h *VulnerabilityHandlers) GetVulnerability(w http.ResponseWriter, r *http.
...
+
+		if errors.As(err, &mr) {
+			return api.WriteJSON(w, mr.status, api.APIError{Error: mr.Error()})
+		} else {
+			return api.WriteJSON(w, http.StatusInternalServerError, api.APIError{Error: err.Error()})
+		}
+	}
+	isUpdated, errUpdate := h.vulnerabilityService.UpdateVulnerabilityComment(vulnID, vulnCommentRequest.Comment)
+	if errUpdate != nil {
+		if errors.Is(errUpdate, sql.ErrNoRows) {
```
grez-lucas, line 138: Good use of error checking for status code :)

pkg/api/api.go
grez-lucas, line 1: Good job keeping endpoints RESTful :+1:

pkg/interfaces/storage.go
```diff
@@ -54,4 +54,7 @@ type IStorage interface {
 	ScanScheduleEnableJob(string, bool, int) error
 	GetHostVulnerabilityTrends(hostID int, timePeriodFilter domain.TimePeriodFilter, severityFilters []string) ([]domain.ServiceTimePeriod, error)
 	GetRapporteursAndHostAliasByScanID(scanID uuid.UUID) ([]*domain.Rapporteur, string, error)
+	UpdateVulnerabilityComment(ID int, comment string) (bool, error)
+	DeleteVulnerabilityComment(ID int) (bool, error)
+	NotExistComment(ID int) (bool, error)
```
grez-lucas: This NotExistsComment method is a good idea for a helper function, there are two things which come to mind which may have been improvements design wise:

1. Renaming the method to `HasComment(vulnID int)`, which avoids funny double negations such as `!NotExistsComment` (harder to read) in `pkg/services/vulnerability.go`
2. (Optional) Choosing not to expose this method via interface, and keep the logic for checking if a comment exists or not inside the storage layer, returning a custom error if the comment does not exist.
- This point is just another option to avoid keeping our storage interface very large, your approach is fine too!

pkg/services/vulnerability.go
```diff
@@ -153,3 +154,22 @@ func calculateVulnerabiltyAgeDays(timePublished time.Time) int {
 	durationInDaysFloat := time.Since(timePublished).Hours() / 24
 	return int(durationInDaysFloat)
 }
+
+func (s *VulnerabilityService) CreateVulnerabilityComment(vulnID int, vulnComment string) (bool, error) {
+	notExists, errSelect := s.storage.NotExistComment(vulnID)
+	if !notExists {
+		return false, customerrors.ErrCommentExists
```
grez-lucas, line 161: This is great!

pkg/handlers/vulnerability.go
grez-lucas: Also, I feel you should've checked here if `errCreate` is `customerrors.ErrCommentExists` and return a different response code in that case. Since that technically would not be an Internal Server Error.

grez-lucas (changes requested):
Very good PR! :tada: Please do the following and we should be good to go:

- Improve error handling at the handler level with suggested reviews
- Check and refactor minor design flaws if they make sense to you
- Add some unit tests for your new `pkg/services/vulnerability.go` methods in `pkg/services/vulnerability_test.go`  ;)

pkg/interfaces/storage.go
cristianoCatolico: I choose to keep my option 😄

pkg/services/vulnerability_test.go
```diff
@@ -302,3 +302,61 @@ func Test_GetVulnerability_ServiceNotFound(t *testing.T) {
...
+
+func Test_CreateVulnerabilityComment_ErrorHasComment(t *testing.T) {
+	mockStore := &mocks.MockStorage{}
+	vulnService := NewVulnerabilityService(mockStore)
+	created, err := vulnService.CreateVulnerabilityComment(1, "test")
+
+	assert.Error(t, err)
+	assert.Equal(t, false, created)
+	assert.ErrorIs(t, err, customerrors.ErrCommentExists)
+}
```
grez-lucas, line 314: This test really left me wondering how we're getting an `ErrCommentExists` without explicitly declaring the return of the mock storage's `storage.HasComment()` method.

Next time, it'd be a better practice to explicitly state the return of the  `storage.HasComment()` method within your test to improve your code's readability. 

I understand now it's because your default mock: 

```go
func (m *MockStorage) HasComment(ID int) (bool, error) {
	if m.MockHasComment != nil {
		return m.MockHasComment(ID)
	}
	return true, nil
}
```

Returns `true, nil` as default. While this is ok, it would be more idiomatic to make your mock return `false, nil`, because the [Zero value for booleans is false](https://yourbasic.org/golang/default-zero-value/).

pkg/mocks/storage.go
grez-lucas, line 1: Some boolean based methods return `true, nil` as default. While this is ok, it would be more **idiomatic** to make your mocks return `false, nil`, because the [Zero value for booleans is false](https://yourbasic.org/golang/default-zero-value/), and that would better reflect your code's behavior.

grez-lucas (approved):
Good work, there are some minor improvements that could be made regarding code readability with mocks and tests. 

Feel free to merge after reading them. Also, good work on updating documentation!👍

PR #99, opened 2025-03-18
Adding logs and correcting verify registration

cristianoCatolico:


pkg/services/auth.go
grez-lucas, line 1: Good work improving logs and cleaning up methods :)

pkg/api/api.go
```diff
@@ -65,7 +65,7 @@ func (s *APIServer) Init() error {
 	router.HandleFunc("POST /api/forgot-password", makeHTTPHandlerFunc(s.authHandlers.ForgotPassword))
 	router.HandleFunc("POST /api/change-password", makeHTTPHandlerFunc(s.authHandlers.ChangePassword))
 	router.HandleFunc("POST /api/users", makeHTTPHandlerFunc(s.authHandlers.RegisterUser))
-	router.HandleFunc("POST /api/users/{id}/verify-email", makeHTTPHandlerFunc(s.authHandlers.VerifyEmail))
+	router.HandleFunc("GET /api/users", makeHTTPHandlerFunc(s.authHandlers.VerifyEmailOnTemplate))
```
grez-lucas: I feel this new api route name is unclear. A better name to keep things RESTful could be something like:

```suggestion
router.HandleFunc("GET /api/users/verify", makeHTTPHandlerFunc(s.authHandlers.VerifyEmailOnTemplate))
```

pkg/interfaces/auth.go
```diff
@@ -26,4 +26,5 @@ type IAuthHandlers interface {
 	VerifyEmail(w http.ResponseWriter, req *http.Request) error
 	ChangePassword(writer http.ResponseWriter, request *http.Request) error
 	WithAuth(endpoint http.HandlerFunc, functionName string) http.HandlerFunc
+	VerifyEmailOnTemplate(w http.ResponseWriter, req *http.Request) error
```
grez-lucas: Please correct me if I'm wrong. Are you using `VerifyEmail` anymore? Since you removed it from the api.go calls, I think this isn't the case. 

If I'm correct, then you should rename `VerifyEmailOnTemplate` to `VerifyEmail` and remove the old `VerifyEmail` method, to make your code clearer.

grez-lucas (commented):
Good work overall. Some minor improvements can be made to method naming. Please consider these suggestions and add logic for the email template URL as we discussed yesterday afternoon. 👍🏻👍🏻

grez-lucas (approved):
Good stuff!

PR #102, opened 2025-03-21
Hotfix/kt 393

cristianoCatolico:


cmd/migrations/migrations/000025_add_scheduled_to_enum_scan.down.sql
grez-lucas, line 1: Good work writing good migrations for this change!

pkg/customerrors/scan.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 var ErrScanNotFound = errors.New("scan not found")
+var ErrScanHostFKNotFound = errors.New("insert or update on table scans violates foreign key constraint scans_host_id_fkey")
```
grez-lucas, line 11: Smart workaround for detecting this error

pkg/storage/scan.go
```diff
@@ -190,13 +196,13 @@ func (s *PostgreSQLStore) GetScans(tenantID string) ([]*domain.ScanSummary, erro
       COALESCE(A.medium, 0) AS medium,
       COALESCE(A.high, 0) AS high,
       COALESCE(A.critical, 0) AS critical
-   FROM  scans S
+   FROM  (SELECT * FROM scans WHERE status!=$2) S
    INNER JOIN hosts H ON S.host_id = H.id
    LEFT JOIN aggregated_vulnerabilities A ON S.id = A.scan_id
    WHERE S.tenant_id = $1
    ORDER BY S.started_at DESC`
```
grez-lucas: Performance wise, I think it might be better to leave this conditional in the WHERE clause:

 ```suggestion
   FROM scans  S
   INNER JOIN hosts H ON S.host_id = H.id
   LEFT JOIN aggregated_vulnerabilities A ON S.id = A.scan_id
   WHERE S.tenant_id = $1 AND S.status != $2
   ORDER BY S.started_at DESC`
```

pkg/storage/scan.go
grez-lucas, line 1: This is a good change! It kind of concerns me now that `storage.GetScans` doesn't get ALL scans, but scans which are not scheduled. It's just a naming thing, perhaps a better name for this method could be `GetCurrentScans` or something of that matter.

grez-lucas (approved):
Solid changes @cristianoCatolico ! Good work. There is only a minor performance improvement with the SQL query and a minor naming concern for `GetScans` which I'd like you to look into before merging. Testing seemed ok and code quality looks clean!

PR #105, opened 2025-03-25
Websocket for Scan

cristianoCatolico:


pkg/config/config.go
```diff
@@ -122,6 +126,7 @@ func load() *Config {
 		FromEmail: fetchEnvOrPanic("SMTP_FROM_EMAIL"),
 		Password:  fetchEnvOrPanic("SMTP_PASS"),
 	}
+	cfg.Websocket = struct{ IntervalScanRefresh string }{IntervalScanRefresh: fetchEnv("WS_SCAN_INTERVAL_NOTIFICATION", "60")}
```
grez-lucas, line 129: Nice use of `config.go` for this!

pkg/handlers/wsclient.go
grez-lucas, line 1: Nice use of GoLang Docs for functions, methods and struct fields in this file. I assume some of it must be inspired from tutorials, but overall it makes readablity much easier :+1:

pkg/handlers/wsclient.go
```diff
@@ -0,0 +1,158 @@
...
+var (
+	// pongWait is how long we will await a pong response from client
+	pongWait = 10 * time.Second
+	// pingInterval has to be less than pongWait, We cant multiply by 0.9 to get 90% of time
+	// Because that can make decimals, so instead *9 / 10 to get 90%
+	// The reason why it has to be less than PingRequency is becuase otherwise it will send a new Ping before getting response
+	pingInterval = (pongWait * 9) / 10
+)
+
+// NewClient is used to initialize a new Client with all required values initialized
```
grez-lucas: GoLang Doc Audit: comment on exported function NewWsClient should be of the form "NewWsClient ..." (ST1020)

```suggestion
// NewWsClient is used to initialize a new Client with all required values initialized
```

pkg/handlers/wsserver.go
```diff
@@ -0,0 +1,110 @@
...
+	server := &WsServer{
+		cfg:         config.LoadConfig(),
+		clients:     make(WsClientList),
+		handlers:    make(map[string]EventHandler),
+		scanService: scanService,
+	}
+	server.setupEventHandlers()
+	return server
+}
+func (m *WsServer) setupEventHandlers() {
```
grez-lucas: Ideally the receiver name on the same type `WsServer` throughout this file should always be the same to improve readability:

```suggestion
func (ws *WsServer) setupEventHandlers() {
```

pkg/api/api.go
grez-lucas, line 1: If we're replacing router.HandleFunc("GET /api/scans", this PR should get rid of that route and handler.

cmd/main.go
```diff
@@ -73,6 +73,7 @@ func main() {
 		eventBus)
 	vulnHandlers := handlers.NewVulnerabilityHandlers(vulnService)
 	scanScheduleHandlers := handlers.NewScanScheduleHandlers(scanScheduleService)
+	wsServer := handlers.NewWsHandlers(scanService)
```
grez-lucas: This raises some eyebrows, our wsServer being a WsHandler gives bad code smell

grez-lucas (changes requested):
Overall, this looks like quite a clean implementation of WS for replacing the `GET api/scans` route! :tada: 

However, there is a design decision here that raises some eyebrows. :eyes: 

Notice how you've defined interfaces and implemented a _handler_ for a WebSocket _server_.  Then, in `main.go` you are **injecting the WS Server into the API Server struct**. There are a couple of pros and cons for this design:

#### Pros :heavy_check_mark: 
* **Centralized Management:** All of our server logic resides in the API server
* **Cohesion:** If the WS server is tightly coupled with the API server (e.g., they share a lot of state or closely interact), it can make sense to group them. (perhaps the case with middleware, but not so true)
#### Cons :x: 

* **Complexity:** It can make the API server struct more complex and harder to maintain.
* **Responsibility:** Violates the Single Responsibility Principle (SRP) by giving the API server more responsibilities. (not just HTTP now, but also WS protocols).

Generally, if the WebSocket server and API server have distinct responsibilities and can operate independently, it is advisable to keep them as separate structs. This approach aligns with the principle of separation of concerns and makes our codebase more modular and maintainable.

Example of separate structs:

```go
type APIServer struct {
    listenAddr         string
    scanHandlers       interfaces.IScanHandlers
    vulnHandlers       interfaces.IVulnerabilityHandlers
    scanScheduleHandlers interfaces.IScanScheduleHandlers
}

type WSServer struct {
    scanHandlers interfaces.IScanHandlers
}

func NewAPIServer(listenAddr string, sHandlers interfaces.IScanHandlers, vHandlers interfaces.IVulnerabilityHandlers, ssHandlers interfaces.IScanScheduleHandlers) *APIServer {
    return &APIServer{
        listenAddr:         listenAddr,
        scanHandlers:       sHandlers,
        vulnHandlers:       vHandlers,
        scanScheduleHandlers: ssHandlers,
    }
}

func NewWSServer(sHandlers interfaces.IScanHandlers) *WSServer {
    return &WSServer{
        scanHanlders: sHandlers,
    }
}
```

In my opinion, this gives us the following pros and cons:

#### Pros :heavy_check_mark: 

* **Separation of Concerns:** Each struct has a single responsibility, making the codebase cleaner and easier to maintain.
* **Flexibility:** Easier to change or replace one component without affecting the other.
* **Reusability:** Components can be reused in different contexts or projects.
* **Extensibility:** Future functionalities that require WebSocket logic can be handled within the WS Handler, such as interactive views in the FE (which will come later).

#### Cons: :x: 

* **Initialization Complexity:** You need to manage the lifecycle and dependencies of multiple structs.
* **Communication Overhead:** If the WS server and API server need to communicate frequently, managing their interaction can add complexity. (I don't think this is the case here)

I feel like the proper approach would be to have separate structs and initialization logic. **Let us discuss this today** and find the most suitable approach. 

## PS: Unused routes and handlers

Since this PR is effectively replacing the current `GET api/scans` route, we should see some deletions on existing interfaces and handlers. This will help clean up the codebase and prevent code bloat.

cmd/main.go
grez-lucas, line 1: The use of goroutines to initialize both the WebSocket server and the API server is a very appropriate change! There are just a couple of improvements which could be made to clean up this implementation and make it more robust and readable.

pkg/api/api.go
```diff
@@ -53,7 +53,7 @@ func NewAPIServer(
 	}
 }
 
-func (s *APIServer) Init() error {
+func (s *APIServer) Init() http.Server {
```
grez-lucas, line 56: Nice refactor to make both server structs return the same type :)

pkg/handlers/wsevent.go
grez-lucas, line 1: This is a good first approach for implementing Event names. For extensibility, consider implementing a handler registration mechanism and ensuring handlers are decoupled and easily integrable with business logic. This will allow us to add new handlers, like one for handling client status updates, without major changes to the core server logic.

grez-lucas (approved):

cmd/main.go
cristianoCatolico: Thanks Lucas!

PR #122, opened 2025-04-21
feat(ScanFailedEvent): Initialize handler and subscribe to event subject

grez-lucas:
This pull request updates the `SetupEventBus` function in `pkg/events/events.go` to handle a new event type, improving the system's ability to process failure scenarios. The logic for handling a `ScanFailedEvent` had been implemented, but not integrated! 😂


cristianoCatolico (approved):
Nice integration when the code flows.

PR #123, opened 2025-04-22
fix: rapporteurs

cristianoCatolico:
This commit has a validation if fields are different from null in order to the edition for rapporteurs and credentials.


pkg/handlers/host.go
```diff
@@ -167,16 +166,16 @@ func (h *HostHandlers) ValidateHost(w http.ResponseWriter, req *http.Request) er
...
 	operatorID := req.Context().Value(middleware.ContextUserID)
 
 	host := domain.NewHost(
-		result.Domain,
-		result.IP,
+		"", //result.Domain,
+		"", //result.IP,
 		tenantID.(string),
 		operatorID.(string),
 		createHostRequest.Name,
```
grez-lucas: This leads to scans not running, since we're using Domain/IP values later as targets for the scan services.

pkg/storage/host.go
```diff
@@ -151,18 +151,30 @@ func (s *PostgreSQLStore) PatchHostByID(h *domain.Host) (*domain.Host, error) {
...
+			SET  rapporteurs=$2, domain=$3, ip=$4, alias=$5
+				WHERE id=$1
+			RETURNING *
+		  `
+		rapporteursJSONB, err := json.Marshal(h.Rapporteurs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal rapporteurs: %w", err)
+		}
+
+		row = tx.QueryRow(query, h.ID, rapporteursJSONB, h.Domain, h.IP, h.Name)
```
grez-lucas, line 175: This looks good 👍🏻

grez-lucas (changes requested):
The changes in `pkg/handlers/host.go` lead to the system not working, since IP and Domain values are required to specify targets for scans in other microservices

grez-lucas (approved):
Looks good 👍🏻

PR #130, opened 2025-05-07
Hotfix/kt 449

cristianoCatolico:
Modification of samples package to generate fake data.

### DIFF
--- a/cmd/sampledata/main.go
+++ b/cmd/sampledata/main.go
@@ -61,7 +61,7 @@ func populateDB(store interfaces.IStorage) {
 	}
 	fmt.Println("Hosts populated successfully")
 
-	if err := populateScans(store, tenants, len(hosts)); err != nil {
+	if err := populateScans(store, tenants, hosts); err != nil {
 		panic(err)
 	}
 	fmt.Println("Scans populated successfully")
@@ -87,23 +87,25 @@ func populateTenants(store interfaces.IStorage) ([]domain.Tenant, error) {
 func populateHosts(store interfaces.IStorage, tenants []domain.Tenant) ([]domain.Host, error) {
 	sampleHosts := samples.SampleHosts(10, tenants)
 
-	for _, host := range sampleHosts {
-		_, err := store.CreateHost(&host)
+	for i, host := range sampleHosts {
+		createdHost, err := store.CreateHost(&host)
 		if err != nil {
 			return nil, fmt.Errorf("error populating host %s: %w", host.Name, err)
 		}
+		sampleHosts[i] = *createdHost
 	}
 	return sampleHosts, nil
 }
 
-func populateScans(store interfaces.IStorage, tenants []domain.Tenant, hostSize int) error {
-	sampleScans := samples.SampleScans(10, tenants, hostSize)
+func populateScans(store interfaces.IStorage, tenants []domain.Tenant, hosts []domain.Host) error {
+	sampleScans := samples.SampleScans(10, tenants, hosts)
 
 	for i, scan := range sampleScans {
 		createdScan, err := store.CreateScan(&scan)
 		if err != nil {
 			return fmt.Errorf("error populating scans: %w", err)
 		}
+		createdScan.EndedAt = scan.EndedAt
 		sampleScans[i] = *createdScan
 	}
 
--- a/cmd/sampledata/main.go
+++ b/cmd/sampledata/main.go
@@ -117,7 +117,7 @@ func populateScans(store interfaces.IStorage, tenants []domain.Tenant, hostSize 
 			nil,
 			scan.ID,
 			enums.StatusCompleted.String(),
-			scan.StartedAt.Add(time.Duration(gofakeit.IntRange(1, 100))),
+			scan.StartedAt.Add(time.Minute*time.Duration(gofakeit.Number(100, 200))),
 		); err != nil {
 			return fmt.Errorf("error updating scan status and ended at: %w", err)
 		}
--- a/cmd/sampledata/main.go
+++ b/cmd/sampledata/main.go
@@ -48,17 +48,19 @@ func Run() {
 func populateDB(store interfaces.IStorage) {
 	fmt.Println("Populating DB with sample data...")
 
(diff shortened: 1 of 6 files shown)


cmd/sampledata/main.go
grez-lucas, line 1: Very good work keeping this top level main.go file concise and easy to read. Changing the signature for the different population funcs to return the created entities and passing them to the next func is very functional and clean design!

cmd/sampledata/main.go
```diff
@@ -115,7 +117,7 @@ func populateScans(store interfaces.IStorage) error {
 			nil,
 			scan.ID,
 			enums.StatusCompleted.String(),
-			scan.StartedAt.Add(time.Minute*5),
+			scan.StartedAt.Add(time.Duration(gofakeit.IntRange(1, 100))),
```
grez-lucas: Nice that you didn't forget about this detail 👍🏻

go.mod
grez-lucas, line 1: Minor audit detail here: don't forget to run `go mod tidy` so that this file and `go.sum` get organized automatically 😉

pkg/samples/host.go
grez-lucas, line 1: Again, great functional programming design here.  Also, good job on keeping private functions lowercase and only keeping public functions (`SampleHosts`) public/exported.

pkg/samples/scan_results.go
```diff
@@ -1,816 +1,213 @@
...
+	weaknessType[2] = enums.WeaknessOther
+	weaknessType[3] = enums.WeaknessBrokenAccessControl
+	weaknessType[4] = enums.WeaknessSecurityLoggingAndMonitoringFailures
+	weaknessType[5] = enums.WeaknessNoInfo
+	exploitableType := make([]enums.ExploitabilityType, 5)
+	exploitableType[0] = enums.ExploitabilityTypeUndefined
+	exploitableType[1] = enums.ExploitabilityTypeUnknown
+	exploitableType[2] = enums.ExploitabilityTypeFunctional
+	exploitableType[3] = enums.ExploitabilityTypeUnproven
+	exploitableType[4] = enums.ExploitabilityTypeProofOfConcept
```
grez-lucas: What you have done here is interesting and a good approach. Depending on if the enum is implemented via `iota` or not, we can also do the following:

```go
var allWeaknessTypes []enums.WeaknessType

for i := 0; i <= int(enums.WeaknessNoInfo); i++ {
   allWeaknessTypes = append(allWeaknessTypes, enums.WeaknessType(i)
}

```

This approach is a little less verbose and **faster to scale/maintain in case we add a new enum.**

This only works for all enums defined with `iota` though, which is a better practice, but not all of our enums in the common package follow that atm (the most recent ones do though!).

[The Iota Docs](https://go.dev/wiki/Iota) briefly show how iota works, in short it's just an incrementing int value.

cmd/sampledata/main.go
grez-lucas: It does seem to generate some weird results though, please consider:

image




```suggestion
			scan.StartedAt.Add(time.Minute * gofakeit.IntRange(1, 100)),
```

pkg/samples/scan_results.go
```diff
@@ -1,816 +1,213 @@
...
+			BaseCVSSScore: math.Trunc(gofakeit.Float64Range(0, 10)*10) / 10,
+			BaseSeverity:  severityType[gofakeit.IntRange(0, 5)],
+			References:    []string{gofakeit.URL()},
+			Exploit: tools.Exploit{
+				Score:          math.Trunc(gofakeit.Float64Range(0, 1)*10) / 10,
+				Exploitability: exploitableType[gofakeit.IntRange(0, len(exploitableType)-1)],
+			},
+		}
+	}
+	return vulns
```
grez-lucas: The vulnerability entity is the most critical in our app, which is why it really matters that we can fake it properly for testing purposes.

Please add the following fake fields for your vulnerabilities (and as many as you can):

- Description
- VendorComments
- RiskScore
- Exploit


The full [Vulnerability Struct definition can be found here](https://github.com/kptm-tools/common/blob/main/common/pkg/results/tools/nmap.go#L52). Please fake as much of it as you can.

grez-lucas (changes requested):
@cristianoCatolico This is a very clean PR and I can tell you've been putting work into it. Congrats 🎉 . There's just one minor detail pending:

1. Enrich the fake vulnerabilities more

---
PD: This improvement you have made to the codebase allowed me to realize that both the `api/reports` and `api/scans` are not filtering by tenant ID 🤭 (so the user can see scans from other hosts right now). I'll add this to the backlog for a later fix

pkg/samples/scans.go
```diff
@@ -9,182 +12,87 @@ import (
...
+		} else if targets[indexTarget] == "domain" {
+			hostValue = gofakeit.DomainName()
+			alias = strings.Split(hostValue, ".")[0]
+			targetType = enums.Domain
+		} else {
+			hostValue = gofakeit.DomainName()
+			alias = strings.Split(hostValue, ".")[0]
+			targetType = enums.Subdomain
+		}
+		hostID := gofakeit.Number(1, hostsSize)
```
grez-lucas: On this line, it appears the logic for selecting a host for the scan is currently picking a random *index* based on the total number of hosts available *across all tenants* in the system.

🚫 This is causing scans to be associated with hosts belonging to tenants other than the one initiating the scan. (What I mentioned in my PD comment). Consequently, users are seeing scans linked to hosts that are not theirs in the scans view.

To fix this, please modify the logic so that:

1.  You first retrieve the list of hosts specifically associated with the authenticated user's tenant.
2.  Then, randomly select a host *from this tenant-specific list*.
3.  The `HostID` and the target information for the scan should be obtained directly from the selected host object *belonging to the correct tenant*.

grez-lucas (changes requested):

pkg/samples/scans.go
grez-lucas, line 1: Very good changes here!

grez-lucas (approved):
Good work! 🏆

PR #132, opened 2025-05-15
Edit host: domain ->IP->Domain

cristianoCatolico:
If we follow domain ->IP->domain, then it remains the IP


main
grez-lucas, line 1: Whoops, we don't want to commit this binary!

pkg/handlers/host.go
```diff
@@ -107,9 +108,19 @@ func (h *HostHandlers) PatchHostByID(w http.ResponseWriter, req *http.Request) e
 	}
 	hostToDB, err := h.constructHostForDB(createHostRequest, req)
 	if err != nil {
-		return err
+		return api.WriteJSON(w, http.StatusInternalServerError, err.Error())
```
grez-lucas: To keep things standard for @diegocarhuavilca , we may want to use our error struct here:

```suggestion
		return api.WriteJSON(w, http.StatusInternalServerError, api.APIError{Error: err.Error()}))
```

Also, it is a good practice to log the error with `slog.Error()` before sending it back to the client.

PD:
As a side note, perhaps its time we write a constructor for this kind of error in the `api` package.. maybe a `api.WriteError(w http.Response, status int, message string)` that writes a JSON version of the error ? This could also allow us to write less _back-end_ specific errors to this struct (e.g: "unkown column 'ip' in 'host', and instead more user-friendly errors (e.g: 'An internal server error occurred').

pkg/handlers/host.go
```diff
@@ -107,9 +108,19 @@ func (h *HostHandlers) PatchHostByID(w http.ResponseWriter, req *http.Request) e
...
+	hostGet, err := h.hostService.GetHostByID(hostToDB.ID)
+	if err != nil {
+		statusCode := http.StatusNotFound
+		return api.WriteJSON(w, statusCode, api.APIError{Error: http.StatusText(statusCode)})
+	}
+	if enums.IP.String() == createHostRequest.ValueType && len(hostGet.Domain) > 0 {
+		hostToDB.Domain = hostGet.Domain
+	} else if enums.Domain.String() == createHostRequest.ValueType && len(hostGet.IP) > 0 {
+		hostToDB.IP = hostGet.IP
+	}
```
grez-lucas: This solves the selection error.

pkg/handlers/host.go
```diff
@@ -107,9 +108,19 @@ func (h *HostHandlers) PatchHostByID(w http.ResponseWriter, req *http.Request) e
...
 	host, err := h.hostService.PatchHostByID(hostToDB)
```
grez-lucas, line 127: Validation must take place in order to avoid letting the user input an invalid IP or Hostname. 

```suggestion
        // Get the value of the host, IP if it's an IP type, Hostname if it's a Domain/Subdomain
        if err := h.hostService.ValidateHost(hostValue); err != nil {
        // Handle the case when the host value (the target) is not valid...
        }
        // Patch the host in the DB if everything's ok
	host, err := h.hostService.PatchHostByID(hostToDB)
```

pkg/handlers/host.go
grez-lucas, line 1: The handler solves the core issue in this task, which is switching the values from IP to Domain in the edit modal:

image
➡️ 
image
✅ 


The problem now is that if the user decides to change the IP to an invalid one:

E.g.
image
➡️ 
image
🚫 

Then a small error is logged in the back-end, but the host was edited. We should **not let the user change the value if it's an invalid value.**

image
I suggest we use the `hostService.ValidateHost()` function for target validation. In technical terms, we just need to **ping the host value (IP or Hostname) and check that it's healthy.**

grez-lucas (changes requested):
This is a step in the right direction. Please read through suggestions and text me if any further clarification is needed 👍🏻

main
cristianoCatolico: Your are right, my fault

pkg/handlers/host.go
cristianoCatolico: Your are right, I have not considered it. Thanks for the suggestion

pkg/handlers/host.go
cristianoCatolico: I found another bug in hostService.ValidateHost, now is working 😄

pkg/handlers/host.go
cristianoCatolico: This constructor will be in the pr of mistakes 🥺

pkg/handlers/host.go
```diff
@@ -107,16 +109,33 @@ func (h *HostHandlers) PatchHostByID(w http.ResponseWriter, req *http.Request) e
...
+	hostGet, err := h.hostService.GetHostByID(hostToDB.ID)
+	if err != nil {
+		statusCode := http.StatusNotFound
+		return api.WriteJSON(w, statusCode, api.APIError{Error: http.StatusText(statusCode)})
+	}
+
+	// Get the value of the host, IP if it's an IP type, Hostname if it's a Domain/Subdomain
+	if errValidation := h.hostService.ValidateHost(createHostRequest.Value); errValidation != nil {
+		// Handle the case when the host value (the target) is not valid
+		slog.Error("Failed to validate host value", slog.String("host_value", createHostRequest.Value))
```
grez-lucas, line 120: Logging the error here would come in handy too 😉

 ```suggestion
		slog.Error("Failed to validate host value", slog.String("host_value", createHostRequest.Value), slog.Any("error", errValidation))
```

File: pkg/handlers/host.go [hunk 5]
```diff
@@ -107,16 +109,33 @@ func (h *HostHandlers) PatchHostByID(w http.ResponseWriter, req *http.Request) e
 	}
 	hostToDB, err := h.constructHostForDB(createHostRequest, req)
 	if err != nil {
-		return err
+		slog.Error("Failed to constructHostForDB", slog.Any("error", err))
+		return api.WriteJSON(w, http.StatusInternalServerError, api.APIError{Error: err.Error()})
 	}
 	hostToDB.ID = id
+	hostGet, err := h.hostService.GetHostByID(hostToDB.ID)
```
grez-lucas: `hostGet` is now declared but not used. It still makes sense to check for host existence in DB, so this block of logic should be changed to:

```suggestion
	_, err := h.hostService.GetHostByID(hostToDB.ID)
```

grez-lucas (approved):
Looks good! 👍🏻 Just some minor audit and logging improvements to make.

Also, don't forget to run `go mod tidy` and commit the `go.sum` file 😉

PR #146, opened 2025-06-26
feat: schema for web vulnerability

grez-lucas:


db/sql/migrations/000024_create_web_vulnerabilities_table.down.sql
grez-lucas, line 1: image


The order of migrations matters! Down migrations are failing because web_vulnerability_details depends on web_vulnerabilities. **Please fix this so that up and down migrations run properly**

grez-lucas: @cristianoCatolico

grez-lucas: Also, please rename your migrations and rebase your branch from `dev` so we don't have duplicate ...00024 migrations 🤓 , remember to use `git mv` !

cristianoCatolico: Thanks for the observation, it is rebased and also I test adding file by file with migrate up and down.

PR #147, opened 2025-06-30
Feature/kt 497

cristianoCatolico:
This feature containss the repository and service layer. Also there is a sample script


db/sql/migrations/000026_create_web_vulnerability_details_table.up.sql
```diff
@@ -1,6 +1,6 @@
 -- Migration: 000026_create_web_vulnerability_details_table.up.sql
 CREATE TABLE IF NOT EXISTS web_vulnerability_details (
-    id UUID PRIMARY KEY,
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
     vulnerability_id UUID NOT NULL,
```
grez-lucas: Does this make sense? I understand from your last schema that **we have a strict one-to-one relationship between them**. Because `web_vulnerability_details` stores an `InstanceAlert`.

E.g., we shouldn't have a `web_vulnerability_details` row that has no `web_vulnerabilities` record.

First, I would suggest changing this table name to `web_vulnerability_intance` for clarity.

Then, I would make sure this `vulnerablity_id` references the `web_vulnerabilities_table` and NOT the base table `vulnerabilities`.

```sql
CREATE TABLE "vulnerabilities" (
  "id" uuid PRIMARY KEY,
  "title" varchar(512) NOT NULL
  -- ... other common fields
);

CREATE TABLE "web_vulnerabilities" (
  "vulnerability_id" uuid PRIMARY KEY, -- This is both the PK and FK
  "solution_advice" text,
  "reference" text,
  CONSTRAINT "fk_web_vulns_base" 
    FOREIGN KEY ("vulnerability_id") 
    REFERENCES "vulnerabilities"("id")
    ON DELETE CASCADE
);

CREATE TABLE "web_vulnerability_details" (
  "id" uuid PRIMARY KEY,
  "vulnerability_id" uuid NOT NULL, -- This links to the table above
  "uri" text NOT NULL,
  "param" text,
  "evidence" text,
  CONSTRAINT "fk_web_details_to_web_vuln" 
    FOREIGN KEY ("vulnerability_id") 
    REFERENCES "web_vulnerabilities"("vulnerability_id") -- References table above
    ON DELETE CASCADE
);
```

grez-lucas: These are the benefits of this approach:

* **Logical Purity:** It correctly models the relationship. A detail record is a detail of a web vulnerability, not just any vulnerability. This structure is self-documenting.

* **Data Integrity:** This is the most important benefit. It makes it impossible at the database level to accidentally associate a web_vulnerability_detail (like a URI) with a non-web vulnerability (like a network_os_vulnerability). The foreign key constraint guarantees that the parent record exists in the web_vulnerabilities table.

* **Clear Responsibility:** Each table has a distinct role. vulnerabilities defines what a vulnerability is, web_vulnerabilities defines what makes it a web vulnerability, and web_vulnerability_details lists where it was found.

pkg/services/vulnerability.go
```diff
@@ -336,3 +339,115 @@ func (s *VulnerabilityService) GetScanVulnerabilities(ctx context.Context, scanI
...
+					Confidence: 0,
+					CPE:        "",
+				},
+				Product:         "",
+				State:           "",
+				Vulnerabilities: nil,
+			}
+			if vuln.WascID == "45" {
+				portData.Protocol = "nginx"
+			}
```
grez-lucas: This portData generation looks like something we would want to keep at the domain level as a function. E.g.,: A function that gets a web vulnerability and tries to create a `tools.PortData` struct from it.

grez-lucas: Crucial oversight here:

The PortData `ID` Field ([reference here](https://github.com/kptm-tools/common/blob/main/common/pkg/results/tools/nmap.go#L37)) refers to the number of the PORT (e.g., 80, 443) and is used as a PRIMARY KEY when storing a service. make sure we're checking this. 


Please refer to how CreateOrUpdateService assigns parameters here: https://github.com/kptm-tools/core-service/blob/dev/pkg/storage/service.go#L45

db/sampledata/main.go
```diff
@@ -232,6 +239,99 @@ func populateNetworkOSVulnerabilities(
...
+				err = deps.VulnRepo.CreateWebVulnerabilityDetail(ctx, dbVuln.ID, vulnDetail.URI, string(vulnDetail.Method), vulnDetail.Param, vulnDetail.Attack, vulnDetail.Evidence, vulnDetail.OtherInfo)
+				if err != nil {
+					return fmt.Errorf("failed to create networkOSVUlnerability record for OS: %w", err)
+				}
+			}
+		}
+	}
+	return nil
+}
+
```
grez-lucas: Sadly, this is the part of the populate script with lots of duplicated code (it mirrors the service layer Insert). It would be ideal to use the service for all insertions. 

Could we use dependency injection to add the Vulnerability Service to this script?

grez-lucas (changes requested):
This is a great first approach 🥇 ! The ticket is quite big, so it's only natural that there will be observations. I'll try to define an order in which to tackle them.  


## 1. Clean Up Schema

It makes more sense for the table `web_vulnerability_details` to be called `web_vulnerability_instances`. The Foreign Key Constraints can be improved as well.

[The comments in this PR Review  illustrate this point a bit better](https://github.com/kptm-tools/core-service/pull/147/files?new_files_changed=true#rundefined)

### ADR

We must document the database design patterns we're following, and this is a great place to start. Please write an ADR on the **Class Table Inheritance** pattern we are using for web vulnerabilities.

Make sure to pull the latest dev changes, since we have one ADR there in `/docs/adr/`.

## 2. Manage Business Logic 

We must manage business Logic better at the **Service Layer.** In particular, you must consider:

1. The scan runs and finishes correctly
1. Nmap will probably have ran first and inserted a service associated to port 80 and port 443 **for that `scan_id`**
2. Our logic must account for this and link web vulnerabilities to these pre-existing services.
3. Our logic must also account for cases in which the services don't previously exist. 

## 3. Sample Script - Avoid Duplicated Code

The `samples` `populate` script seems to be duplicating code from the service layer (at many points, not just your service). This is a technical debt problem we must fix.

For now, please try to inject the _Service_ in charge of inserting web vulnerabilities, so we can avoid having to change code at two places later. 

I will add a backlog ticket to fix the rest of that script's functions during out technical debt cycles.

db/sql/migrations/000026_create_web_vulnerability_details_table.up.sql
cristianoCatolico: You are right, it is better putting as your correction, the benefits are many. Thanks

pkg/services/vulnerability.go
cristianoCatolico: Now I understand better thanks

db/sampledata/main.go
cristianoCatolico: For the two vulnerabilities network and webscan, I am using service injection.

docs/adr/ADR-003.md
```diff
@@ -0,0 +1,29 @@
...
+**Date:** 2025-07-07
+
+## 💭 Context
+
+When implementing database, and considering that we are using class inheritance pattern for "is-a" relationship problem. For example: a `WebVulnerability` is a type of `Vulnerability`, and a `NetworkVulnerability` is also a type of `Vulnerability`. They share common traits such as a title, a description, and a severity, but each has unique fields. For instance, a web vulnerability has a `wasc_id`, while a network vulnerability needs a `cve_id`.
+Additionally, there is a another table of `WebVulnerabilityInstances` that is direct specialization of `WebVulnerability`.
+One can model, that `WebVulnerabilityInstances` is related to `Vulnerability` for a better query of data, how is when having this nesting situation of inheritance?
+
+## 💡 Decision
+When having nested inheritance is better to be related to the first parent, otherwise, there can be that other child that has a parent in common have data that does not correspond to its domain. For instance having that `WebVulnerabilityInstances` is related to `Vulnerability` will cause that `NetworkVulnerability` can have `WebVulnerabilityInstances` where does not belong to the domain.
```
grez-lucas: This is a great explanation of a complex topic; I'm happy to help polish the English a bit to ensure it's easy for everyone to follow

pkg/services/vulnerability.go
```diff
@@ -336,3 +339,133 @@ func (s *VulnerabilityService) GetScanVulnerabilities(ctx context.Context, scanI
...
+			service, errGetExistence := s.serviceRepo.GetServiceByScanHostPortProtocol(ctx, scan.HostID, scan.ID, int32(portData.ID), portData.Protocol)
+			if errGetExistence != nil && !errors.Is(errGetExistence, customerrors.ErrServiceNotFound) {
+				return fmt.Errorf("failed to get service by scan host port and protocol: %w", errGetExistence)
+			}
+			if service == nil {
+				service, err = s.serviceRepo.CreateOrUpdateService(ctx, scan.HostID, scan.ID, portData)
+				if err != nil {
+					return fmt.Errorf("failed to create or update service: %w", err)
+				}
+			}
```
grez-lucas, line 366: Nice

pkg/services/vulnerability_test.go
```diff
@@ -436,3 +440,192 @@ func TestVulnerabilityService_GetScanVulnerabilities_NotFound(t *testing.T) {
 	require.Nil(t, vulns)
 	assert.ErrorIs(t, err, customerrors.ErrScanNotFound)
 }
+
+// -- Tests for CreateNetworkOSVulnerabilities ---
```
grez-lucas:
```suggestion
// -- Tests for CreateWebScanVulnerabilities ---
```

pkg/services/vulnerability_test.go
```diff
@@ -436,3 +440,192 @@ func TestVulnerabilityService_GetScanVulnerabilities_NotFound(t *testing.T) {
...
+		return nil
+	}
+	// Act
+	err := service.InsertWebScanVulnerabilities(ctx, scanID, webScanResult)
+
+	// Assert
+	assert.NoError(t, err)
+}
+
+func TestVulnerabilityService_CreateWebScanVulnerabilitiesWhenServiceNOTExists_Success(t *testing.T) {
```
grez-lucas, line 543: Nice test!

grez-lucas (changes requested):
Great job @cristianoCatolico ! It seems you covered the main observations from the last PR. The missing point right now are unit tests.


## Unit Tests are a little Poor....

Just two success unit tests in `pkg/services/vulnerability_test.go` does not seem very thorough.... **What happens when our repository stuff fails ?!** 🤯 

Please add unit tests considering cases where calls to different repository methods fail, this is why we have mocks!

Also, please add unit tests for the `GetWebScanVulnerabilities(ctx context.Context, scanID uuid.UUID) ([]tools.WebVulnerability, error)` function you wrote as well. 

These tests may sound boring to implement, but they will help us track bugs quickly when a part of our system fails. 

With improvements to the unit tests for the `VulnerabilityService`, we should be good to merge!

docs/adr/ADR-003.md
cristianoCatolico: Thanks Lucas!

pkg/services/vulnerability_test.go
cristianoCatolico: Thanks!

pkg/mocks/storage/cwe.go
```diff
@@ -51,6 +51,6 @@ func (m *MockCWERepo) CreateOrUpdateCWEFromWebVulnerability(ctx context.Context,
 	if m.MockCreateOrUpdateCWEFromWebVulnerability != nil {
 		return m.MockCreateOrUpdateCWEFromWebVulnerability(ctx, vuln)
 	}
-	panic(fmt.Sprintf("MockCWERepo: method CreateOrUpdateCWEFromWebVulnerability called but not implemented for test: %s", ctx.Value(testutil.TestNameKey)))
-
+	return nil, nil
+	//panic(fmt.Sprintf("MockCWERepo: method CreateOrUpdateCWEFromWebVulnerability called but not implemented for test: %s", ctx.Value(testutil.TestNameKey)))
```
grez-lucas: ⚠️ ⚠️

grez-lucas (changes requested):

pkg/mocks/storage/cwe.go
cristianoCatolico: resolved

grez-lucas (approved):
Nice! 👍🏻

PR #148, opened 2025-07-01
Kt 498: Fix SQL join condition to correctly fetch CWE details in vulnerability queries

luizarnoldch:
This PR fixes a critical issue in the GetVulnerabilityWithDetailsByID SQL query where the cwe_details table was incorrectly joined on a non-existent column cwe_id from the cve_details table. The join condition has been corrected to join vulnerabilities.cwe_id directly with cwe_details.cwe_id, matching the actual database schema.

In addition, the corresponding Go structs and query scanning code have been updated to reflect the fixed query. This ensures that vulnerability detail queries return complete and accurate data including CWE information.

Changes included:

SQL query fix: corrected join condition in vulnerability detail fetch query.
Regenerated Go SQL boilerplate code for the updated query.
Added error handling improvements where applicable.
Verified and tested the fix to prevent SQL errors and data omissions.
This resolves the backend failure encountered when retrieving vulnerability details and improves data completeness and API stability.


pkg/dto/response.go
```diff
@@ -248,12 +248,45 @@ type ScanVulnerabilityDetailResponse struct {
 	VendorComments []tools.VendorComment `json:"vendor_comments"`
 	References     []string              `json:"references"`
 
+	Metrics []CVSSMetric `json:"metrics"`
+
+	CWERemediation []CWERemediation `json:"remediation"`
```
grez-lucas: It would make sense for this to be plural, wouldn't it?

```suggestion
	CWERemediations []CWERemediation `json:"remediation"`
```

pkg/dto/response.go
```diff
@@ -248,12 +248,45 @@ type ScanVulnerabilityDetailResponse struct {
...
+
+	EPSSScore      *float64   `json:"epss_score,omitempty"`
+	EPSSPercentile *float64   `json:"epss_percentile,omitempty"`
+	EPSSDate       *time.Time `json:"epss_date,omitempty"`
```
grez-lucas, line 257: Good work here, this makes the JSON cleaner and more compact when these are absent. It clearly signals that these fields are optional and only appear when relevant.

db/sql/queries/vulnerability.sql
```diff
@@ -72,7 +72,7 @@ FROM
 LEFT JOIN
   cve_details AS cve ON v.cve_id = cve.cve_id
 LEFT JOIN
-  cwe_details AS cwe ON v.cwe_id = cve.cwe_id
+  cwe_details AS cwe ON v.cwe_id = cwe.cwe_id
```
grez-lucas, line 75: nice catch!

docs/adr/ADR-001_Fix_join_condition_in_vulnerability_details_query.md
grez-lucas, line 1: Good first use of an ADR! It does sound like Alternative 1 is the only real alternative though. 

For future ADRs, I'd rather have them be about more significant **architectural changes**, not fixes (because the alternatives are just fixing it 🤣 ). 

For example, the use of pointers and `omitempty` in your DTO structs makes for a valuable DTO, since:
1. it had to balance the use of two patterns in order to design clean and predictable DTOs for our API.
2. it can be used for later reference by the team, so we can follow similar practices when choosing one pattern or another 😉 


I suggest **deleting this ADR**, and changing it's content for the DTO struct patterns you chose to use, and why:
* Pointers for null Values (Explicit Absence)
* Using omitempty (Implicit Absence)

Mentioning pros and cons, and when to use one over the other (or a combination of both as you just did!)

pkg/dto/response.go
grez-lucas, line 1: Good use of pointers and `omitempty` here. It is **crucial** that whatever choice we make, you document properly. We must clearly communicate the contract to the API consumers.

This means (using **OpenAPI/Swagger**):

* **Documenting Pointers:** A pointer field like `PrivilegesRequired *string` should be documented in OpenAPI with `type: string` and `nullable: true`. This explicitly tells the client to expect either a string or `null`.

* **Documenting `omitempty`:** A field with `omitempty` is simply documented as an optional field (i.e., not listed in the `required` section of the schema). The client's code must be able to handle the key not being present in the JSON response.



Since we're not using `swaggo` yet, you'll have to [update the OpenAPI documentation repository](https://github.com/kptm-tools/kptm-docs)'s `swagger.yaml` file.

grez-lucas (changes requested):
Good changes, and good work! Not only did you provide a cleaner DTO, but also were proactive about fixing that JOIN issue. 

Here is a summary of my observations:

## Minor fixes

Some minor fixes regarding plural naming for fields which are slices. Please clarify if the use of a singular field name is intentional if that's the case.

## ADR

I love the initiative you took, but ADRs are not for fixes. Instead, they should aim to document key architectural/design decisions or patterns we're using in development. I've added a suggestion for an ADR based on the contents of this PR.

## Documentation

I need you to manually document the changes made to the VulnerabilityDetailDTO.

Please refer to the `swagger.yaml` file from [our docs repo](https://github.com/kptm-tools/kptm-docs)

PR #149, opened 2025-07-02
fix: status

cristianoCatolico:


grez-lucas (commented):
Nice catch! @diegocarhuavilca please let us know if this solves the issue, and approve if that's the case.

diegocarhuavilca (approved):

PR #151, opened 2025-07-07
Kt 504

luizarnoldch:
1. Create the endpoint GET /api/scans/{id}/assets
2. Create the swagger and bruno endpoint and test it.
3. Generate the swagger doc


pkg/storage/scan.go
```diff
@@ -85,10 +86,59 @@ func (r *ScanRepo) GetScanByID(ctx context.Context, scanID uuid.UUID) (*domain.S
 	if err != nil {
 		return nil, err
 	}
+
```
cristianoCatolico, line 88: We should use slog debug for this information

cristianoCatolico (changes requested):
Overall great job with documentation and implementation of the feature. Just one observation of using slug. debug for one part that you were using fmt.print. Just that change for me.

pkg/handlers/scan.go
```diff
@@ -142,9 +142,64 @@ func (h *ScanHandlers) CreateScan(w http.ResponseWriter, req *http.Request) erro
...
+	if len(rawResults) == 0 {
+		slog.Warn("No assets found for scan",
+			slog.String("scan_id", scanID.String()))
+		return api.WriteJSON(w, http.StatusNotFound, api.APIError{Error: fmt.Sprintf("Scan Assets for the ID %s not found", scanID.String())})
+	}
+
+	response := dto.ConvertScanOSandServicesResultToResponse(rawResults)
+
+	return api.WriteJSON(w, http.StatusOK, response)
+}
```
grez-lucas, line 198: Very clean handler! 🏆

pkg/storage/scan.go
grez-lucas: Whoops! Looks like you forgot to delete these debug `Println` , since we're logging at the handler level, I think we could just delete these :)

grez-lucas (changes requested):
Solid implementation! 🏆 

The service doesn't have too much logic to test, but perhaps we could add some **handler unit tests** for the `func (h *ScanHandlers) GetScanAssetsByID`, since we have a little bit of business logic in there. 

## What to test? 🤔 
- Request parsing
- Error handling from services
- Business logic checks (`scan.Status` is "Completed")
- Response formatting (`http.StatusNotFound` if the service returns an empty slice of assets, etc.

pkg/storage/scan.go
luizarnoldch: Great Catch. I already updated.

grez-lucas (approved):
Nice addition of tests! I hope it helped you familiarize yourself with `httptest` and some of our mock interfaces :)

PR #154, opened 2025-07-14
Kt 506

luizarnoldch:


db/sql/queries/vulnerability.sql
```diff
@@ -101,24 +101,24 @@ ORDER BY
...
-  CAST(COALESCE(SUM(CASE WHEN severity = 'Low' THEN 1 ELSE 0 END), 0) AS integer) AS low_vulnerabilities,
-  CAST(COALESCE(SUM(CASE WHEN severity = 'Medium' THEN 1 ELSE 0 END), 0) AS integer) as medium_vulnerabilities,
-  CAST(COALESCE(SUM(CASE WHEN severity = 'High' THEN 1 ELSE 0 END), 0) AS integer) as high_vulnerabilities,
-  CAST(COALESCE(SUM(CASE WHEN severity = 'Critical' THEN 1 ELSE 0 END), 0) AS integer) AS critical_vulnerabilities
+  CAST(COALESCE(SUM(CASE WHEN severity = 'UNKNOWN' THEN 1 ELSE 0 END), 0) AS integer) AS unknown_vulnerabilities,
+  CAST(COALESCE(SUM(CASE WHEN severity = 'NONE' THEN 1 ELSE 0 END), 0) AS integer) AS none_vulnerabilities,
+  CAST(COALESCE(SUM(CASE WHEN severity = 'LOW' THEN 1 ELSE 0 END), 0) AS integer) AS low_vulnerabilities,
+  CAST(COALESCE(SUM(CASE WHEN severity = 'MEDIUM' THEN 1 ELSE 0 END), 0) AS integer) as medium_vulnerabilities,
+  CAST(COALESCE(SUM(CASE WHEN severity = 'HIGH' THEN 1 ELSE 0 END), 0) AS integer) as high_vulnerabilities,
+  CAST(COALESCE(SUM(CASE WHEN severity = 'CRITICAL' THEN 1 ELSE 0 END), 0) AS integer) AS critical_vulnerabilities
```
grez-lucas, line 121: Nice fix! 👍🏻

pkg/dto/response.go
```diff
@@ -243,11 +243,17 @@ type ScanAssetsResponse struct {
...
 	Alias                string                  `json:"alias"`
+	IPAddress            string                  `json:"ip_address"`
+	OSName               string                  `json:"os_name"`
+	OSType               string                  `json:"os_type"`
 	TotalVulnerabilities int                     `json:"total_vulnerabilities"`
 	SeverityCounts       tools.SeverityCounts    `json:"severity_counts"`
 	Vulnerabilities      []ScanVulnerabilityItem `json:"vulnerabilities"`
+	CWERemediations      []CWERemediation        `json:"remediation"`
+	References           []string                `json:"references"`
 }
```
grez-lucas: This is a good point of debate. This particular struct is being used for ALL kinds of scan vulnerability Items, but it contains specific _Operating System_ fields, such as **OSName** and **OSType**. 

Perhaps it would be wise to rename this DTO struct and its comment to something like `DetectedOS` or `OSScanResult` (because it contains metadata about the OS, and vulnerabilities).


## OS Accuracy Field ✨ 

Could we please add an `OSAccuracy` field here, which references the value of the `domain.OperatingSystem.Accuracy` which is an `int32` ? 

This field would be extremely valuable, because it tells the analyst how certain we are about this OS finding (it's always 0-100).

pkg/handlers/scan.go
```diff
@@ -500,6 +503,163 @@ func (h *ScanHandlers) GetScanVulnerabilities(w http.ResponseWriter, r *http.Req
...
+			slog.Any("error", err),
+		)
+		return api.WriteJSON(w, http.StatusInternalServerError, api.APIError{
+			Error: http.StatusText(http.StatusInternalServerError),
+		})
+	}
+
+	if scan.Status != "Completed" {
+		return api.WriteJSON(w, http.StatusConflict, api.APIError{Error: "Scan status not completed"})
+	}
```
grez-lucas, line 552: Very clean up to this point! 😄

pkg/handlers/scan.go
```diff
@@ -500,6 +503,163 @@ func (h *ScanHandlers) GetScanVulnerabilities(w http.ResponseWriter, r *http.Req
...
+		return api.WriteJSON(w, http.StatusInternalServerError, api.APIError{
+			Error: http.StatusText(http.StatusInternalServerError),
+		})
+	}
+
+	if len(vulnerabilities) == 0 {
+		slog.Warn("No vulnerabilities found for scan",
+			slog.String("scan_id", scanID.String()))
+		return api.WriteJSON(w, http.StatusNotFound, api.APIError{Error: fmt.Sprintf("Scan Vulnerabilities for the ScanID %s not found", scanID.String())})
+	}
```
grez-lucas: Consider when there's no vulnerabilities for the scan's OS, but we actually detected the Operating System. 

Returning a `404 Not Found` would be a bit confusing because the resource the user is requesting—the operating system asset—_does_ exist

## ✅ Best Practice: `200 OK` with an Empty Array
We should return the DTO containing the Operating System metadata (OSName, OSType, etc..) but:

- Empty `ScanVulnerabilityItem` Slice
- Zero value severity counts
- No remediations 

Instead of returning a 404 not found

pkg/handlers/scan_test.go
```diff
@@ -101,3 +181,224 @@ func TestScanHandlers_GetScanAssetsByID(t *testing.T) {
...
+			emailService:        &mock_services.MockEmailService{},
+			eventBus:            &events.NatsEventBus{},
+			r: func() *http.Request {
+				r := httptest.NewRequest("GET", "/api/scans/"+scanID.String()+"/assets", nil)
+				r.SetPathValue("id", scanID.String())
+				r = r.WithContext(context.WithValue(r.Context(), middleware.ContextRoles, []domain.Role{domain.RoleAdmin}))
+				return r
+			}(),
+			wantStatus: http.StatusNotFound,
+		},
```
grez-lucas: This test should now be refactored to return a `200 OK` with empty vulnerabilities, remediations and Zero value severity counts

pkg/services/vulnerability.go
```diff
@@ -147,6 +147,26 @@ func (s *VulnerabilityService) CreateNetworkOSVulnerabilities(
...
+	for _, vuln := range vulns {
+		detail, err := s.GetScanVulnerabilityDetail(ctx, vuln.NetworkVulnerability.VulnerabilityID)
+		if err != nil {
+			return nil, err
+		}
+		vulnDetails = append(vulnDetails, *detail)
+	}
+	return vulnDetails, nil
+}
+
```
grez-lucas: This method now gets ScanVulnerabilityDetails structs, but specifically for the vulnerabilities associated to an Operating System. 

Perhaps it is time we rename it? What do you think?

grez-lucas (changes requested):
Great implementation @luizarnoldch ! Love the clean coding patterns. 

## Observations 👀 

There's a particular case with the handler when no vulnerabilities are found. Since this EP's resource is not strictly the vulnerabilities, but also some OS metadata, it would be better to return empty slices instead of a 404 error. 

## General Comments 🗨️ 

As you might have noticed, there's several methods and functions which are becoming more granular in this PR. E.g.,

* Getting vulnerabilities -> Getting OS Vulnerabilities
* Getting ScanVulnerabilityDetails -> Getting VulnerabilityDetails for a Scan's Detected Operating System.

For easier maintenance later on, feel free to rename any methods when you feel they are becoming more granular like this. It will drastically help us improve code readability !

pkg/dto/response.go
luizarnoldch: I think i will review it because i found some other struct when this is part of. I do not if i can correct this or create a new one without break the others.

pkg/handlers/scan_test.go
luizarnoldch: Fixed

pkg/handlers/scan.go
luizarnoldch: Fixed

pkg/handlers/scan.go
luizarnoldch: Thanks!

db/sql/queries/vulnerability.sql
luizarnoldch: Thanks!

pkg/dto/response.go
grez-lucas: I suggest we create a new one! Which would include these new fields. You are **correct** that this would break the others.

grez-lucas: ## Acciones

1. Conseguir TODAS las vulnerabilidades de un scan <- `ScanVulnerabilityItemsResponse`
2. Conseguir las vulnerabilides NetworkOS del Sistema Operativo detectado en el Scan **<- We are here**
3. Conseguir las vulnerabilidades NetworkOS de algun servicio detectado en el Scan
4. (Pendiente) Conseguir las vulnerabilidades Web de algun servicio detectado en el Scan

luizarnoldch: Fixed

pkg/services/vulnerability.go
luizarnoldch: Fixed

grez-lucas (approved):
Good work @luizarnoldch ! 🏆

PR #155, opened 2025-07-18
Kt 506

luizarnoldch:
Update the Assets response, version was empty because it was not complete by the CPE. Name is fixed to only show the name of the operation system and not the fullname.


luizarnoldch:
Closed because i added unnecesary last commit. I need to make a new branch to make my changes.

PR #156, opened 2025-07-18
update the response from assets to define the name and fullname with …

luizarnoldch:
Short Update to send the short name and full name about the operation system and define the exact version based on the CPE.

### DIFF
--- a/pkg/dto/response.go
+++ b/pkg/dto/response.go
@@ -1,6 +1,7 @@
 package dto
 
 import (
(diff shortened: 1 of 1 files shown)


grez-lucas (approved):
👍🏻

grez-lucas:
@diegocarhuavilca The accuracy for a scan's operating system can be now found at the 
- `GET /api/scans/{id}/assets` EP

PR #159, opened 2025-07-25
remove wasc detail

cristianoCatolico:
This PR contains remove of table wasc_details and all code that use the table, thus it is removed test related to wasc_repo


grez-lucas (approved):
Nice use of migrations to achieve this! This PR will help us remove bloat from our application 😄

PR #162, opened 2025-07-30
Bug/kt 530 bug main dashboards fail to load due to outdated queries

grez-lucas:
# ✨ Project Polish: Database, Auth & Stability Improvements! ✨

This Pull Request helps fix faulty dashboards and give the system an overall much needed optimization.

## 🧽 Database Spring Cleaning & Optimizations

* Completely removed the old WASC data model and all its related database queries and code. This simplifies our system and keeps it lean! 🧹

## Query Power-Ups! 🚀

* Database queries are now more consistent (using cwe_id for better matching).

* Added smart filters to prevent errors when gathering data, ensuring smoother information flow. ✅

## Tidier Code: 🏷️

* Removed unnecessary "public" prefixes from database references, making our SQL code cleaner.

## 🔒 Smarter & More Flexible Authentication

* Fixed a bug where authentication issuer was being checked incorrectly

## 🚦 Smoother Error Handling & Debugging

* Instead of flagging hosts with no scans as errors or warnings, the system now quietly skips them. This leads to much cleaner logs and more efficient processing overall! 📊


db/sql/queries/vulnerability.sql
cristianoCatolico, line 1: I have also detected that in my pr and cleaned it

pkg/handlers/auth.go
cristianoCatolico, line 1: clean implementation

pkg/storage/scan.go
cristianoCatolico, line 1: excelent detection of bug

cristianoCatolico (changes requested):
The execution of this pr is returning  null in some fields, 
Screenshot 2025-07-30 at 9 46 20 PM

Can you check please, I do not think that should return null or yes?

grez-lucas:
@cristianoCatolico It's correct, since that's how we differentiate between:

- A month with 0 vulnerabilities found

- a month with no data (no scans for that month)

db/sql/queries/vulnerability.sql
grez-lucas: Nice! I think your pr only fixes the vulnerabilities.sql queries that mentioned wasc_id, this one fixes generated files, and the scans.sql queries as well. Since it's more complete, let's stick with this one!

